PR #84, opened 2013-12-05
Allow properties in TranslateableProperties to be updated

victorsavu3:
adds observers on __Translation in order to update the translation


jamesarosen:
v4.0 alpha (see #213) has live bindings on locale.

PR #99, opened 2014-05-08
Made a ember object of i18n

lrdevries:
by doing this its possible to reopen the object and change the translation


jamesarosen:
v4.0 alpha (see #213) has live bindings on locale.

PR #135, opened 2014-09-01
Bound translateable element attributes

hontas:
Working example here:
http://emberjs.jsbin.com/dopoz/2/edit

Using "data-bindtranslateattr-xxx" as to not confilct with regular bound attributes.


owilliams:
Looks like this will become invalid if/once #182 is merged.

jamesarosen:
Closing now that #182 is merged. Please reopen if there's something we should reconsider here.

Thanks for contributing!

PR #147, opened 2014-09-30
{{t}} helper should update multiple bound properties.

GoodForOneFare:
The first observed property triggers a rerender, effectively removing it
from the DOM until the render loop runs.  The second observer examines
the view, finds that view.$() returns null, and assumes it's safe to
destroy itself.

To work around the above, keep the observer around when the view is in
"preRender" state.


jamesarosen:
I don't think this is an issue in v4+. If you want to submit a pull request to patch v3.1, I'll gladly get it merged and released.

PR #148, opened 2014-10-02
Package up for the rails asset pipeline.

tmcgilchrist:
We're using this for some of our apps at work and wanted it packaged up as a ruby gem.
I'm not sure if you want to add another package format to your project but thought it might be interesting.

Cheers


jamesarosen:
v4.0 (see #211) of this library will be only available as an Ember-CLI addon and will be written using ES2015. I am interested to see how other projects make Ember-CLI addons work with the Rails asset pipeline. If there are some things we can do in the project to make that easier, please let me know.

pjmorse:
We are developing our app using Ember-CLI in a distinct subdirectory within our Rails project. We have a custom Rake task to build the app and "export" it into the Rails asset directory for production deploys; it gets included in its own context but the asset pipeline just copies it, and doesn't concatenate or minify it. So we're making Ember-CLI addons work by bypassing the asset pipeline. The downside is that there is some duplication of vendor scripts (e.g. Bootstrap or jQuery UI components) between the Rails side and the Ember side, but that seems like a relatively small price to pay, and it keeps both Ember and Rails on their own happy paths.

Issue #154, opened 2014-10-17
Ember i18n lazy lookup


sbounmy:
Hello,

I found Rails lazy lookup great and looking to implementing it for ember i18n.
Basically when the translation key starts with a dot, it will prefix the key with the template path/name e.g:

``` js
// in templates/users/new.hbs
{{t ('.title') }} // will lookup users.new.title
Sign up!
// in templates/users/_form.hbs
{{t ('.password.label') }} // will lookup users._form.password.label
Enter your password

// Our i18n would look like :
Ember.I18n.translations = {
  users: {
    new: {
      title: 'Sign up!'
    },
    _form: {
     password: {
       label: 'Enter your password'
     }
   }
  }
}
```

Any guideline on how I could fetch the template name/path where I am calling `{{ t 'my_key'}}` translation helper ?
Note I'm using ember-cli and es6 transpiler

jamesarosen:
Do you want to try this again on v4.0.0?

sbounmy:
I would love to, I have a working prototype on my app however I've found myself monkey patching ember-i18n to make it work.
@jamesarosen which version of ```ember``` is ```ember-i18n``` running against ? I'm still stuck at 1.8.X because of ```ember-easyform```atm

jamesarosen:
The [wiki](https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Installation#note) suggests you should be on Ember-I18n v3.x.

cibernox:
About this topic, I found it very interesting, specially in the context of translating error messages in forms.

That's why if first made this PR #230 to ember-i18n (released in v4.1.0) to enable me to build some mechanism to translate my errors in the context of the current route with an addon https://github.com/cibernox/ember-i18n-errors

This although my addon is focused in errors, this pattern of fallbacks based in the route hierarchy is useful in any context, and ember-i18n can eventually embrace it and absorb my addon.

In my mind the ideal syntax for "route-scoped" translations could be just implicitly do it for those keys starting with a dot, like rails-i18n does.  

That way while `{{t 'some.trans.lation'}}` is an absolute path, `{{t '.trans.lation'}}` is a relative path to the current route.

Being the current route name `parent.child.grandchild`, the relative path would try to find the better translation following this fallback chain:

```rb
parent.child.grandchild.trans.lation
parent.child.trans.lation
parent.trans.lation
trans.lation # or maybe application.trans.lation because the top-level route is the Application route.
```

[...]

PR #155, opened 2014-10-19
Lazy helper lookups

RSSchermer:
Pull request implementing lazy lookups for the `{{t}}` helper, as per https://github.com/jamesarosen/ember-i18n/issues/154.

There is good news and there is bad news. The good news is that `options.data.view.get('renderedName')` seems to work pretty well for determining the template name.

* It is only defined for top level templates, not for other views or components.
* When including a partial view, the `renderedName` remains the name of the including view, not the included partial view. This could actually be considered useful.
* I was afraid that `renderedName` might actually be based on the route name, rather than the template name, but this is not true. When explicitly changing the template rendered by a route, the `renderedName` becomes the name of that template:

```javascript
// routes/resource/index.js
import Ember from 'ember';

export default Ember.Route.extend({
  renderTemplate: function() {
    this.render('test');
  }
});
// The renderedName of this route's view will now be 'test', not 'resource.index'
```

However, there's also bad news. When using the `{{t}}` inside another view helper, `options.data.view` will be the view of that helper. As mentioned `renderedName` is only defined for top level route views, so it will return `undefined`:

```handlebars
<!-- test.hbs, translations hash contains a key for 'test.resource-link' -->
{{#link-to 'resource.index'}}
  {{t '.resource-link'}} <!-- Wont work, 'renderedName' is undefined -->
{{/link-to}}
```

One way around this is to loop up through the parent views of `options.data.view` until a view is found that does have a `renderedName` property. There is however a further problem with this: Ember's public API seems to treat all the top level route template views as one single view with `renderedName = 'application'`:

```javascript
function getRenderedName(view) {
    var renderedName = view.get('renderedName');

    while (!renderedName && view) {
      view = view.get('parentView');
      renderedName = view.get('renderedName');
    }

    return renderedName; // Always returns 'application'
}
```

However, Ember's private API does respect the route template views as separate parent views:

```javascript
function getRenderedName(view) {
    var renderedName = view.get('renderedName');

    while (!renderedName && view) {
      view = view.get('_parentView'); // Prefixed with underscore
      renderedName = view.get('renderedName');
    }

    return renderedName; // Will correctly return 'test' when the {{t}} helper is used in templates/test.hbs
} 
```

This seems to work perfectly. I've tested this with a largish Ember app I'm working on and didn't get any unexpected results. It does however use Ember's private API. I'm not sure why this difference between `parentView` and `_parentView` exists. Maybe we could ask the person responsible for this API, but I have no idea who that is.

On the subject of testing: I added 2 simple unit tests. I think this may also benefit from integration tests to verify that using it in a certain template produces the right `renderedName`, but I don't know how I'd go about that with the current test setup. Maybe along the lines of how ember cli addons do it with a dummy app, making ember cli a dependency. I think that might be a pretty big refactor though, not sure if you'd wanna go there.

This doesn't work with translatable attributes yet, but I think I could probably use that same looping up through `_parentView` to get that working.

### DIFF
--- a/lib/i18n.js
+++ b/lib/i18n.js
@@ -215,9 +215,21 @@
       warn("Ember.I18n t helper called with unquoted key: %@. In the future, this will be treated as a bound property, not a string literal.".fmt(key));
     }
 
+    if (key.charAt(0) === '.') {
+      var view = data.view;
+      var parentView = view.get('_parentView');
+
+      while (!view.get('renderedName') && parentView) {
+        view = parentView;
+        parentView = view.get('_parentView');
+      }
+
+      key = view.get('renderedName') + key;
+    }
+
     var translationView = TranslationView.create({
       context: attrs,
-      translationKey: key.charAt(0) === '.' ? data.view.get('renderedName') + key : key,
+      translationKey: key,
       wrappingTagName: tagName
     });
 
--- a/lib/i18n.js
+++ b/lib/i18n.js
@@ -103,6 +103,17 @@
     }
   }
 
+  function getRenderedName(view) {
+    var renderedName = view.get('renderedName');
+
+    while (!renderedName && view) {
+      view = view.get('_parentView');
+      renderedName = view.get('renderedName');
+    }
+
+    return renderedName;
+  }
+
   I18n = Ember.Evented.apply({
     compile: compileTemplate,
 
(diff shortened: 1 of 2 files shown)


jamesarosen:
I think there are enough unresolved issues on #154 that we should close this. We have it to refer to once we resolve that RFC.

RSSchermer:
Do you want me to keep the fork alive for future reference or can I delete it?

jamesarosen:
I saved your branch as [rsschermer/lazy-helper-lookups](https://github.com/jamesarosen/ember-i18n/tree/rsschermer/lazy-helper-lookups), so you can delete.

PR #180, opened 2014-12-19
Escape interpolations by default

opichals:
This is in order to get the expected handlebars-like mustache
escaping behavior. Unescaped interpolation made possible via the
common tripple curlies mustache.

### DIFF
--- a/lib/i18n.js
+++ b/lib/i18n.js
@@ -58,11 +58,17 @@
     }
   }
 
+  var escapeExpression = EmHandlebars.Utils.escapeExpression;
+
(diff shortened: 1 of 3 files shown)


opichals:
@jamesarosen Great! Thank you!

Could you please add a 3.0.0-beta3 tag so that bower treats it as greater than 3.0.0-beta2 (the already present `3.0.0-beta.3` does not do the expected thing).

jamesarosen:
Oh, I understand now. I missed the parenthetical point in your earlier post.

PR #182, opened 2014-12-24
Drop support for {{translateAttr}}

jamesarosen:
Dropping for a variety of reasons:

 * It doesn't  support bound values, which confuses users
 * Users can achieve the same effect with `TranslateableAttributes` or by creating translated properties on the controller or component
 * It doesn't play nicely with the current HTMLbars-Handlebars compatibility layer

See https://github.com/emberjs/ember.js/issues/9961


beastmst:
hi, can give me an example how can I do with my template?
previously i use 
<a href="#"  data-toggle="tooltip" data-placement="bottom" {{translateAttr title="nav.aboutus"}}>

as this is gone, how should i code it to work?

jamesarosen:
How about using a sub-expression?
```handlebars
<a href="#" title='{{unbound (t "nav.aboutus")}}'>
```

beastmst:
thanks, it works :)

shareit:
I changed the line

<span class="czn-icon czn-settings-icon active" {{translateAttr title="lblSoftProofing"}}></span>

into 

<span class="czn-icon czn-settings-icon active" title='{{unbound (t "lblSoftProofing")}}'></span>

and the following error message appears:
Uncaught TypeError: Cannot read property 'indexOf' of undefined.
Please advice.

jamesarosen:
> Uncaught TypeError: Cannot read property 'indexOf' of undefined. Please advice.

Can you provide a stack trace? Do you know where the `indexOf` is getting called?

shareit:
It seems that the error appear in the bower_components/ember/ember.js at the following line:

var firstDotIndexCache  = new Cache(1000, function(key) { return key.indexOf('.');             });
In this case, the "key" variable is undefined.

Here is the full stack trace as it is displayed in the console:
(anonymous function) @ ember.js:16711
Cache.get @ ember.js:11380
(anonymous function) @ ember.js:16714
Cache.get  @ ember.js:11380
getFirstKey  @ ember.js:16755
Stream.get  @ ember.js:18633
CoreView.extend.getStream  @ ember.js:43846
unboundHelper  @ ember.js:8804

ncoquelet:
Hi,

I got the same error as you @shareit !
after having removed the brackets, all works fine !!

replace
```
<span class="czn-icon czn-settings-icon active" title='{{unbound (t "lblTitle")}}'></span>
```
by 

```
<span class="czn-icon czn-settings-icon active" title='{{unbound t "lblTitle"}}'></span>
```

jamesarosen:
Thanks, folks :)

Issue #188, opened 2015-02-04
Possible memory leak in TranslateableProperties


nullrocket:
Since observers are added manually they don't get torn down when the view that the mixin is added to is destroyed, took a while to find where I was leaking!  I fixed it locally by adding a willDestroyElement and willClearRender to the mixin  and moving the anonymous function passed to eachTranslatedAttribute to a method on the Mixin, but I am unsure it that is the "way" to do it.

```javascript 
TranslateableProperties: Ember.Mixin.create({
      init: function() {
        var result = this._super.apply(this, arguments);
        eachTranslatedAttribute(this, function(attribute, translation) {
          this.addObserver(attribute + 'Translation', this, function(){
            set(this, attribute, I18n.t(this.get(attribute + 'Translation')));
          });
          set(this, attribute, translation);
        });

        return result;
      }
    }),

**Comment by jamesarosen** [comment]
I guess we'll need to call `removeObserver` in `willDestroy`. The problem is that `removeObserver` needs the `attribute` *and* the observer callback. That means we need to store those off somewhere :/

```js
_addTranslationObservers: function() {
  eachTranslatedAttribute(this, function(attribute, translation) {
    var callback = function() {
      set(this, attribute, I18n.t(this.get(attribute + 'Translation')));
    };
    this.addObserver(attribute + 'Translation', this, callback);
    this.set('__' + attribute + 'TranslationObserver', callback);
    set(this, attribute, translation);
  });
},

_removeTranslationObservers: function() {
  eachTranslatedAttribute(this, function(attribute, translation) {
    var callback = this.get('__' + attribute + 'TranslationObserver');
    this.removeObserver(attribute + 'Translation', this, callback);
  });
}
```

nullrocket:
Here is what I ended up with.
```javascript
    TranslateableProperties: Ember.Mixin.create({
        _translationObserver :  function(sender, key){
          //  set(this, attribute, I18n.t(this.get(attribute + 'Translation')));
            set(this, key, I18n.t(this.get(key + 'Translation')));
        },
      init: function() {
        var result = this._super.apply(this, arguments);
        var self = this;
        eachTranslatedAttribute(this, function(attribute, translation) {
          this.addObserver(attribute + 'Translation', this,self._translationObserver);
          set(this, attribute, translation);
        });

        return result;
      },

        willDestroy : function(){
            var self= this;
            this._super();

            eachTranslatedAttribute(this, function(attribute, translation) {
[...]

PR #189, opened 2015-02-05
Clean up translated property observers on destroy.

nullrocket:
#188

I couldn't find an event that matched .on('destroy') at http://emberjs.com/api/classes/Ember.View.html  both **willDestroyElement** and **willClearRender** are needed though.

### DIFF
--- a/lib/i18n.js
+++ b/lib/i18n.js
@@ -109,7 +109,7 @@
     },
 
     TranslateableProperties: Ember.Mixin.create({
-        _translationObserver :  function(sender, key, value){
+        _translationObserver :  function(sender, key){
            set(this, key, I18n.t(this.get(key + 'Translation')));
         },
       init: function() {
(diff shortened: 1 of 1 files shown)


jamesarosen:
Nice work. I fixed the test, did a little whitespace cleanup, and committed as 524837cb711c3c204d7a12aff3241f48a9616e5f.

lib/i18n.js
```diff
@@ -109,17 +109,27 @@
     },
 
     TranslateableProperties: Ember.Mixin.create({
+        _translationObserver :  function(sender, key){
```
jamesarosen: This `key` already has the `'Translation'` suffix since that's the property we're observing.

nullrocket:
It was late I gave up for the night after travis failed. I appreciate the fix to the PR, I was in a hurry and just edited in Github since my local repo has too many changes and would have caused a huge pain to do a pull from :)

jamesarosen:
I accidentally left a typo in when I pushed. I have a fix for the new JSHint error ready as soon as I get to work.

Thanks for digging in and finding this!

PR #191, opened 2015-02-09
Updating jshint version squashes npm warning

pjmorse:
npm (which gets auto-corrected to "nom", which I almost don't want to fix) doesn't like packages without repository fields in their package.json. jshint 2.0.x didn't have that field; it's been added since then, so an update gets rid of that annoying (but harmless) warning.

Reference: http://stackoverflow.com/q/16827858/306084


jamesarosen:
+1

PR #149, opened 2014-10-06
Simple ember-cli support.

chrmod:
It is very simple ember-cli support that uses addon mechanism. 

When used in ember-cli project it:
* provide a generator to install cldr and ember-i18n as bower components
* imports cldr and ember-cli to application build
* setup configuration

To use it in ember-cli project:
```
npm install ember-i18n --save-dev
ember generate ember-i18n
```

Implements issue: https://github.com/jamesarosen/ember-i18n/issues/129

PR #192, opened 2015-02-09
Minimal ember-cli addon

pjmorse:
This is a continuation of @chrmod 's work in PR #149. I've rebased it on master and removed the configuration block (because the specified flags are no longer needed). I'm fiddling with installing it in a fork of our app to confirm that it works as advertised, but it seems to install properly.

To try it out: check out this branch into a directory, then run

> `$ npm install /path/to/ember-i18n/ --save-dev`
> `$ ember generate ember-i18n`

It will grab ember-i18n from Bower, so (perhaps ironically) version 2.9 will be installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,20 @@ Example using pluralization in the template:
 
 Depending on the locale there could be up to 6 plural forms used, namely: 'zero', 'one', 'two', 'few', 'many', 'other'.
 
+### Using with Ember-cli
+
+Install ember-i18n as node module:
+```
+npm install ember-i18n --save-dev
+```
+
+Run generator to fetch dependencies:
+```
+ember generate ember-i18n
+```
+
+That's it.
+
 ### Limitations
 
  * There is no way to pass interpolations to attribute translations. I can't
--- a/lib/ember-addon/blueprints/ember-i18n/index.js
+++ b/lib/ember-addon/blueprints/ember-i18n/index.js
@@ -1,0 +1,13 @@
+'use strict';
+
+module.exports = {
+  normalizeEntityName: function() {
+    // this prevents an error when the entityName is
+    // not specified (since that doesn't actually matter
+    // to us
+  },
+
+  afterInstall: function() {
+    return this.addBowerPackageToProject('ember-i18n');
+  }
+};
--- a/lib/ember-addon/index.js
+++ b/lib/ember-addon/index.js
@@ -9,17 +9,6 @@ module.exports = {
     return path.join(__dirname, 'blueprints');
   },
 
-  config: function(env, config) {
-    return {
-      EmberENV: {
-        FEATURES: {
-          I18N_COMPILE_WITHOUT_HANDLEBARS: true,
-          I18N_TRANSLATE_HELPER_SPAN: false
-        }
-      }
-    }
-  },
-
   included: function(app) {
     this._super.included(app);
 
--- a/lib/ember-addon/index.js
+++ b/lib/ember-addon/index.js
@@ -1,0 +1,35 @@
+'use strict';
+
+var path = require('path');
+
+module.exports = {
+  name: 'ember-i18n',
+
+  blueprintsPath: function() {
+    return path.join(__dirname, 'blueprints');
+  },
+
+  config: function(env, config) {
+    return {
+      EmberENV: {
+        FEATURES: {
+          I18N_COMPILE_WITHOUT_HANDLEBARS: true,
+          I18N_TRANSLATE_HELPER_SPAN: false
+        }
+      }
+    }
+  },
+
+  included: function(app) {
+    this._super.included(app);
+
+    this.app.import(app.bowerDirectory + '/ember-i18n/lib/i18n.js', {
+      exports: {
+        'ember-i18n': [
+          'default'
+        ]
+      }
+    });
+    this.app.import(app.bowerDirectory + '/ember-i18n/lib/i18n-plurals.js');
+  }
+};
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
   "keywords": [
     "ember",
     "internationalization",
-    "i18n"
+    "i18n",
+    "ember-addon"
   ],
   "author": "James A. Rosen <james.a.rosen@gmail.com>",
   "repository": {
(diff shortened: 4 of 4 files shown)


locks:
I don't understand the `$ ember generate ember-i18n` command. If you want to install the dependencies, shouldn't you be using one of the `$ ember install` commands?

pjmorse:
Probably; I picked that up directly from @chrmod. I'm still getting comfortable with ember-cli so any direction towards the happy path is eagerly accepted.

locks:
Grep in http://www.ember-cli.com/ for `install:addon` to get an idea where you can go with it.

pjmorse:
`ember install:addon ember-i18n` seems to work (with me using the path to my local dev copy) - it installs the package in my `bower_modules` directory. But:

* I don't have `Ember.I18n` available in my app until I add `app.import('bower_components/ember-I18n/lib/i18n.js');` to my `Brocfile.js`. This doesn't seem right; I expected Bower packages to be brought in more or less automatically. I wonder if this is because ember-i18n uses `lib/i18n.js` as its entry point instead of `index.js`?

* It's working fine on the console, but getting translations loaded has been another hurdle; where's the appropriate place to put them so they're included in the build? (I tested by defining the translations in the console.)

pjmorse:
I'm seeing ember-cli/ember-cli#3142 as well, but from the bug and associated PR that doesn't seem to be our problem.

locks:
1. I thought https://github.com/jamesarosen/ember-i18n/pull/192/files#diff-155f468e3315048285ed69a72c22e8e8R15 covered this so you didn't have to import it yourself. Hm.

2. Good question. Haven't looked too much how `ember-i18n` and `ember-cli` interact on that front. We should figure this out and add an `ember generate translation` command post-PR.

pjmorse:
Yeah, I wonder if there's something not quite right about that import line that makes the import process not work. Or maybe it doesn't get called?

The add-translation-file process sounds like a blueprint to me.

jamesarosen:
Would it help to *first* convert this to ES-2015?

pjmorse:
Very much, I think. Unfortunately, the day I had to spend on this is now gone, so if I'm going to push it forward any more I'm going to have to look for bits and pieces of time (and get un-blocked).

As this stands - reiterating the comments above - it *works* but all it does is make `ember addons:install ember-i18n` functionally identical to `bower install ember-i18n`. To really make it useful as an ember-cli addon I think it needs...

- [ ] To be automatically available in the build without being explicitly added to the `Brocfile.js`.
- [ ] To have a default blueprint which drops a stub translation file at an appropriate place in the project to also be included in the build.

I'm stuck on both of those and would love some direct hints from people who understand ember-cli addons better than I do.

mjperrone:
+1 but I don't have the knowledge to help =[

mansona:
@pjmorse when you say "add it to the brocfile" do you mean the addon's brocfile or the consuming app's brocfile? You can add things to an addon's brocfile and they will get executed when the consuming app included the addon.

Have you used ```ember generate addon``` to build this addon? if not I would highly recommend it, it sets up all the necessary hooks. I could spend a bit of time with you on this on a Google Hangout sometime to walk through this sort of stuff if you were interested?

pjmorse:
@mansona I picked up the work @chrmod did in #149 rather than starting from scratch, so no, I didn't use `ember generate addon`. But that's a good idea; when I get a few more minutes to work on this I'll give it a try. I didn't know that about the Brocfile - I was adding to the application Brocfile to include the library, but it sounds like adding a Brocfile here would do the trick. If it does, we'd only need to figure out generating a stub translations file for the blueprint.

mansona:
cool thing, if you want any more help on this just let me know. 

one of our devs has done a cool thing with blueprints where it runs a particular process after npm install an ember-cli addon. I don't know much about exactly what it's doing but here is an example https://github.com/Blooie/ember-cli-notifications/blob/master/blueprints/ember-cli-notifications/index.js

might be worth looking into for your "stub translations file"

chrmod:
This work looks good in my opinion.

There are couple core ember-cli concepts that are missing on this discussion:
* `ember install:addon ember-i18n` in equalivent of `npm install ember-i18n`, not 'bower install ember-i18n`. It is used to add addon code, not your bower dependencies. 
* `ember generate ember-i18n` is a convention used among addon builders that run the code which prepare project to use the addon. In our case it must only install ember-i18n as bower component so it can be used during build. Note that ember-i18n must first be installed as node module, so the generate will have ember-i18n task.
* global objects like `Ember.I18n` are considered as anti pattern in ember-cli world. To access I18n object you should use `import i18n from 'ember-i18n';`, but ATM ember-i18n is not exported as a module so this wont work. In this line you can see that Ember.I18n is in fact exported as a property on Ember object https://github.com/jamesarosen/ember-i18n/blob/master/lib/i18n.js#L145
* `ember generate addon` should be used to create standalone addon repositiories - not like in our case to add addon repositiory to an existing project.

Finally ember-i18n does not suggest you the one way to load your translations. You can do it in very different ways depending on your project requirements. Those I find the typical approaches:
* ajax get static/generated translation file from your API server
* ajax get static translation file from your ember-cli build (you must add those files to Brocfile.js be part of your build and accessible on certain locations)
* keep translation files as a part of your app (it's just a hash, right?) so you can import them as any other javascript module when it's needed in you app code

All of above you should mix to create best usability of you app. The code doing that should probably be an initalizer.

jamesarosen:
I've spent the last couple days trying to get Ember-I18n to look more like a *native* Ember-CLI addon. A lot has to change and I'm not quite there. I see two options:

 1. merge this and call it v3.0 -- the v3 series will have Ember-CLI support using the legacy global-style API
 2. merge this and continue working on transforming it into ES2015, etc.

pjmorse:
Yeah, the more I learn the more I think ember-i18n needs some structural revamping to *really* work with ember-cli, as you're saying.

That said, I guess I don't understand the difference between your two options: they both seem to boil down to "merge this and keep working."

mansona:
I would really vote for a "merge this and keep working" approach, this will allow us to get some clarity on the current position of the project and really plan the next couple of releases effectively. The sooner we can create a v3.0 (100% stable or not) the better

jamesarosen:
I've started the ES2015 port in the [cli branch](https://github.com/jamesarosen/ember-i18n/tree/cli).

PR #195, opened 2015-02-10
Start testing against Ember 1.10

jamesarosen:
@pjmorse @chrmod 

 * add `ember-template-compiler.js` from Ember
 * conditionally include `ember-template-compiler.js` when testing Ember >= 1.10


jamesarosen:
I considered pulling in `ember-template-compiler.js` by adding `bower install` to the dependencies process and getting it from `bower_components/ember/ember-template-compiler.js`. The only problem is that this library tests against and declares compatibility with versions of Ember that don't include that file.

We might be able to do
```json
dependencies: {
  "ember": "~1"
},
devDependencies: {
  "ember": "~1.10"
}
```

I'm open to discussing this further.

jamesarosen:
@pjmorse @locks @chrmod I'm going to merge this as-is for beta.4. I'm happy to revisit the *how do I get `ember-template-compiler.js`* problem later.

PR #196, opened 2015-02-13
Support passing the context to missingMessage

grantovich:
With this change, the `Ember.I18n.missingMessage` function receives a second argument that is the context passed to the `t` function. This way, in your missing message, you can spit out the data that was supposed to be interpolated in the translation.

I moved some logic around to get access to the context – I'm pretty sure the result is equivalent, and it still passes all the tests. One question I had is whether it's intentional that `missingMessage` is not documented in the README (perhaps because 3.0 is still "beta"). If desired, I'd be happy to add this documentation.

### DIFF
--- a/spec/tSpec.js
+++ b/spec/tSpec.js
@@ -59,17 +59,27 @@ describe('Ember.I18n.t', function() {
     expect(Ember.I18n.t('nothing.here')).to.equal('Missing translation: nothing.here');
   });
 
-  it('warns with a custom message', function() {
-    Ember.I18n.missingMessage = function(key) { return "there.is." + key + ".to.see"; };
-    expect(Ember.I18n.t('nothing.here')).to.equal('there.is.nothing.here.to.see');
-  });
+  describe('missing message', function(){
+    beforeEach(function() {
+      this.oldMissingMessage = Ember.I18n.missingMessage;
+    });
 
-  it('warns with a custom message that includes the passed context', function() {
-    Ember.I18n.missingMessage = function(key, context) {
-      var values = Object.keys(context).map(function(key) { return context[key]; });
-      return key + ',' + (values.join(','));
-    };
-    expect(Ember.I18n.t('foo', { arg1: 'bar', arg2: 'qux' })).to.equal('foo,bar,qux');
+    afterEach(function() {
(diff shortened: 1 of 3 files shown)


jamesarosen:
I'm for passing the context to `missingMessage`. `Ember.Evented#trigger` accepts multiple arguments; I suggest adding the context there as well.

I'd love to have `I18n.missingMessage` documented as part of the public API.

I don't love having the *missing* logic in `I18n.t`, though I don't have a firm suggestion for a better home. Maybe a private `findOrDefineTemplate(key, context)`, which calls `findTemplate` and falls back to defining the missing template.

grantovich:
> `Ember.Evented#trigger` accepts multiple arguments; I suggest adding the context there as well.

Makes sense, I'll do that.

> I'd love to have `I18n.missingMessage` documented as part of the public API.

Okay! I can do it as a separate commit in this PR if that works for you (it seems related enough).

> I don't love having the *missing* logic in `I18n.t`

I agree it's not the best; I'll see what I can do to extract it.

jamesarosen:
We're going to be moving all of the code around in this project to get on ES2015 and provide more native Ember-CLI support. I think the *idea* here is great. It might be good to put this on hold while we reorganize.

grantovich:
@jamesarosen I've made the suggested changes, except for extracting the logic since I don't know how much it would conflict with your reorganization. Are you going to hold off on merging this until after those efforts are done? What is the status/timeline on that?

spec/tSpec.js
```diff
@@ -64,6 +64,14 @@ describe('Ember.I18n.t', function() {
     expect(Ember.I18n.t('nothing.here')).to.equal('there.is.nothing.here.to.see');
   });
 
+  it('warns with a custom message that includes the passed context', function() {
+    Ember.I18n.missingMessage = function(key, context) {
```
jamesarosen: Since this is a global, you'll want to reset it to its default behavior after the spec.

grantovich: Good catch, I didn't think of it since I just copied the above test, which also doesn't reset it. I'll fix that one while I'm in there.

grantovich:
Feedback addressed. Let me know if anything else is needed here.

jamesarosen:
This is fantastic :)

Issue #198, opened 2015-03-04
Upgrade to 1.10 causes TranslateableProperties to require setting property to null


MattNguyen:
In the process of upgrading from 1.09 to 1.10, controllers that included `Ember.I18n.TranslateableProperties` required setting the underlying translateable property to null. 

For example:

```javascript
var ApplicationController = Ember.ObjectController.extend(Ember.I18n.TranslateableProperties, {
  title: null,
  titleTranslation: 'path.to.translation'
});
```

This makes sense since controllers proxy properties to models, and if the model doesn't define the property, then Ember will complain about delegating to the content property of object. 

However, I'm not sure why this worked in 1.09. Any help would be appreciated.

jamesarosen:
Is there something we can do in code to make this work automatically? Or should we just document the requirement?

MattNguyen:
@jamesarosen -  Check out this: https://github.com/MattNguyen/ember-i18n/commit/e53461471a491675d9384e5363e0280958162cec.

Basically, run all the observer initialization stuff in a `scheduleOnce` in the `afterRender` queue. This seems to work for my use case, but the `TranslateableProperties` tests are now broken and I'm not sure why. 

What do you think?

jamesarosen:
To fix the tests, we'll probably have to use more `Ember.run` calls to make sure everything is flushed by the time we're making assertions.

blimmer:
Any updates on this PR? I'm migrating stuff over to 1.11 and this problem has come to bite me several times.

jamesarosen:
@blimmer do you want to try adding `Ember.run` to the tests on top of https://github.com/MattNguyen/ember-i18n/commit/e53461471a491675d9384e5363e0280958162cec ?

blimmer:
I can but probably not this week.

MattNguyen:
Sorry I meant to comment here that I already fixed the tests. They should
be good to go.

MattNguyen:
See: https://github.com/jamesarosen/ember-i18n/pull/199

MattNguyen:
Closing for https://github.com/jamesarosen/ember-i18n/pull/199

PR #199, opened 2015-03-10
Schedule setting up observers in afterRender queue

MattNguyen:
https://github.com/jamesarosen/ember-i18n/issues/198

When including TranslateableProperties in ObjectControllers, setting
properties to their translations fail because the model (ie content)
isn't available. This change schedules this work in the afterRender queue so
that models are guaranteed to be set on the controller.

### DIFF
--- a/spec/translateablePropertiesSpec.js
+++ b/spec/translateablePropertiesSpec.js
@@ -1,16 +1,14 @@
 describe('TranslateableProperties', function() {
 
   it('translates ___Translation attributes on the object', function() {
-    var subject = Ember.Object.extend(Ember.I18n.TranslateableProperties).create({
-      titleTranslation: 'foo.bar'
-    });
+    var klass = Ember.Object.extend(Ember.I18n.TranslateableProperties);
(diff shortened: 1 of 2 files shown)


spec/translateablePropertiesSpec.js
```diff
@@ -1,15 +1,21 @@
 describe('TranslateableProperties', function() {
 
   it('translates ___Translation attributes on the object', function() {
-    var subject = Ember.Object.extend(Ember.I18n.TranslateableProperties).create({
-      titleTranslation: 'foo.bar'
+    var subject;
+    Ember.run(function() {
+      subject = Ember.Object.extend(Ember.I18n.TranslateableProperties).create({
+        titleTranslation: 'foo.bar'
+      });
```
jamesarosen: What do you think of the following?
```js
var klass = Ember.Object.extend(Ember.I18n.TranslateableProperties);
var subject = Ember.run(klass, 'create', { titleTranslation: 'foo.bar' });
```

MattNguyen:
@jamesarosen - updated.

jamesarosen:
+1

Issue #131, opened 2014-08-20
Use registerBoundHelper for {{t}}


jamesarosen:
We can't make the switch until v3 since it's a breaking change for people who use `{{t some.unquoted.key}}`.

I believe this will resolve #47 (`{{t (t "this.translates.to.a.key")}}`), #41 and #113 (`{{t concat("prefix" bound.suffix)}}` or `{{t "prefix" bound.suffix}}`).

Drop support for `I18N_TRANSLATE_HELPER_SPAN` at the same time.

Issue #41, opened 2013-03-08
Using property as translation key, plus scoping


rlivsey:
Currently the `{{t}}` helper is passed a string key, I've recently had the need to translate based on the value of a property (in this case `order`).

For now I've just added a property on the controller for `translatedOrder`:

```
translatedOrder: function(){
    var order = this.get("order");
    return Ember.I18n.t("buttons.order."+order");
}.property("order")
```

That works fine, but nicer would be something like:

```
{{t order scope="buttons.order"}}
```

Obviously we'd then need to differentiate between property keys and string keys...

We could require that string keys are quoted, but that's not very backwards compatible and extra typing for the common case.

We could use a different helper for property translations? `{{pt order}}` or something better!

What do people think?

PR #200, opened 2015-03-10
{{t}} composite keys

chrmod:
Simple feature that allow to build translation keys by passing multiple argument to `{{t}}` helper. Example:
`{{t "foo" "bar"}}` is equal to `{{t "foo.bar"}}`. As helper arguments can be properties on template context you can write something like this: `{{t "foo" someProperty}}`.

This change is a result of previous discussions: https://github.com/jamesarosen/ember-i18n/issues/131 https://github.com/jamesarosen/ember-i18n/pull/113 https://github.com/jamesarosen/ember-i18n/issues/41

### DIFF
--- a/lib/i18n.js
+++ b/lib/i18n.js
@@ -144,8 +144,20 @@
 
   Ember.I18n = I18n;
 
-  EmHandlebars.registerBoundHelper('t', function(key, options) {
-    return new EmHandlebars.SafeString(I18n.t(key, options.hash));
+  EmHandlebars.registerBoundHelper('t', function() {
+    var args = Array.prototype.slice.call(arguments),
+        lastArgument = args[args.length-1],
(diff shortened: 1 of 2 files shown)


chrmod:
Hi @jamesarosen, is there anything we can to do to improve this one?

jamesarosen:
This should be trivial with sub-expressions and a `concat` helper:
```hbs
{{t (concat "foo" someProperty)}}
```

jamesarosen:
I'm all for this in v4.0 -- see #211.

If you think it's worth bringing in to v3.2, I'm on-board. We'll just need to add an example in the README.

jamesarosen:
I think `concat` covers this pretty well. If you want to reopen the discussion, feel free. I'm all ears!

PR #201, opened 2015-03-20
Fix bug with "sticky" missing message context

grantovich:
Fixing a bug that I introduced while adding the "missing message context" feature. The first time `t` was called with a missing key, it would mistakenly cache the context of that call and pass it to all subsequent `missingMessage` calls, even when `t` was later called with the same key but a different context.

This has minor performance implications, since the "missing function" for a given key is now regenerated every time `t` is called with that key. I couldn't immediately think of a better way to fix the problem, and hopefully no one is relying too much on the performance of the missing-translation handler anyway.

### DIFF
--- a/lib/i18n.js
+++ b/lib/i18n.js
@@ -92,7 +92,7 @@
       if (context == null) context = {};
       template = I18n.template(key, get(context, 'count'));
 
-      if (template == null) {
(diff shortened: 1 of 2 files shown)


jamesarosen:
Great catch!

PR #203, opened 2015-03-31
add rtl markers if I18n.rtl is set

craig-day:
Wrap the translated text in directional unicode characters to help with right-to-left text rendering.

### DIFF
--- a/lib/i18n.js
+++ b/lib/i18n.js
@@ -100,7 +100,11 @@
         I18n.trigger('missing', key, context);
       }
 
-      return template(context);
+      if (I18n.rtl) {
+        return "\u202B" + template(context) + "\u202C";
+      } else {
+        return template(context);
+      }
     },
 
     exists: keyExists,
--- a/spec/tSpec.js
+++ b/spec/tSpec.js
@@ -59,6 +59,21 @@ describe('Ember.I18n.t', function() {
     expect(Ember.I18n.t('nothing.here')).to.equal('Missing translation: nothing.here');
   });
 
+  describe('rtl markers', function() {
+    afterEach(function() {
+      Ember.I18n.rtl = undefined;
+    });
+
+    it('leaves markers off by default', function() {
+      expect(Ember.I18n.t('foo.bar')).to.equal('A Foobar');
+    });
+
+    it('adds rtl markers if I18n.rtl is set', function() {
+      Ember.I18n.rtl = true;
+      expect(Ember.I18n.t('foo.bar')).to.equal('\u202BA Foobar\u202C');
+    });
+  });
+
   describe('missing message', function(){
     beforeEach(function() {
       this.oldMissingMessage = Ember.I18n.missingMessage;


jamesarosen:
I'm 100% for supporting RTL. I don't understand browser RTL support very well, though. I had assumed people would solve this problem with CSS:
```css
.my-element { direction: rtl; }
```
but that doesn't really seem to do anything to the text in my tests.

Adding these markers changes the way the text behaves (it right-aligns within inline elements), but doesn't actually *change* the order. That is
```
Hello               [ Hello         ]
\u202BHello\u202C   [         Hello ]
```
Would you mind providing screenshots of some page before and after this change? Just a little block with some *actual* RTL text?

craig-day:
Sure thing.  Some of the things it helps with most are punctuation and parenthesis. If you would like anymore just let me know. 

Notice the question mark here. 



Parenthesis

jamesarosen:
Thanks!

Here are some questions we should probably consider:

 * What happens with interpolations of mixed direction? For example, what if `Ember.I18n.rtl === true` and you interpolate my (English) name into the Hebrew translation of `"Hello, {{name}}"`? How is a Hebrew or Arabic name supposed to appear when embedded in LTR text?
 * Is it conceivable to have the source of the translation strings insert the marker tags? Is that overly burdensome? Would it not work in some edge cases I haven't considered? It has the benefit of keeping the metadata *very close* to the data.
 * Is this better served by way of a *feature* (as you've written) or a *guide*? That is, could this library "support" RTL by having a paragraph of text in the README about how to wrap `I18n.t` with a function that adds these marker characters?

craig-day:
All good questions, here are my opinions on some of them. 

 - English that gets interpolated into the string as a normal English word.  If you take a look at [this website](http://www.pc.co.il/it-news/181795/) you can see what they do with the word Micron, a proper noun. 

 - It is conceivable, but I think it would introduce a lot of work for translation teams that have lots of string in a lot of places. I can't think of any other edge cases. 
 - Similar to the previous point, this sounds reasonable but I can't think of a good reason it couldn't do both.  This functionality will only change if someone explicitly says `I18n.rtl = true`

jamesarosen:
I haven't heard from anyone about this. I'm going to merge it.

craig-day:
@jamesarosen awesome! Thanks for the help.  If you want to toss me on the contributors list, that would be cool!

craig-day:
@jamesarosen how does tagging/versioning work here? Should this go to a beta tag? Just a patch bump, maybe a minor version bump? Can I just bump it, or do you want to take care of that?

jamesarosen:
This should probably be 3.1.0. I haven't written up the release procedure, but it should be something like

 1. Update version number in `package.json`, `bower.json`, and `VERSION`
 1. Commit with the message "vX.Y.Z"
 1. tag that commit "vX.Y.Z"
 1. `git push && git push --tags`
 1. `npm publish` (which I probably have to do)

PR #204, opened 2015-04-07
Do not assume there's prototype extension

kenips:
With EmberCLI 0.2.2+ [ember-disable-prototype-extensions is included by default](http://www.ember-cli.com/#section-1), so we cannot assume them to be there anymore.

### DIFF
--- a/lib/i18n.js
+++ b/lib/i18n.js
@@ -126,18 +126,18 @@
         });
       },
 
-      _scheduleObservers: function() {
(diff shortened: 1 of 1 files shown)


jamesarosen:
I'm +1 on this, but it looks like you'll need to rebase against master.

kenips:
done!

jamesarosen:
+1

PR #206, opened 2015-04-20
Removed script tags from the read me examples

kiwiupover:
I wonder if people might be confused by the metamorph script tags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,9 +88,7 @@ yields
 yields
 ```html
 <h2>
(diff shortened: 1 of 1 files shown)


jamesarosen:
The latest version of Ember-I18n still supports pre-HTMLBars Handlebars. I'm absolutely for this for v4, which will not.

jamesarosen:
See #211

PR #207, opened 2015-04-28
Add translation observers in render queue

MattNguyen:
My use case centers on passing a translateable
property to a component, such as the prompt for a select dropdown.
Putting this work in the `afterRender` queue will not cause the
translations to be loaded before rendering. Hence, this work belongs in
the `render` queue.

### DIFF
--- a/lib/i18n.js
+++ b/lib/i18n.js
@@ -127,7 +127,7 @@
       },
 
       _scheduleObservers: function() {
-        Ember.run.scheduleOnce('afterRender', this, this._addTranslationObservers);
+        Ember.run.scheduleOnce('render', this, this._addTranslationObservers);
       }.on('init'),
 
       _removeTranslationObservers: function (){


jamesarosen:
Can you explain the change from #199?

MattNguyen:
As I mentioned in the description, setting up the observers in the `afterRender` queue doesn't appear correct since translations could load after rendering, which causes `null` translation values in templates. Moving this work to the `render` queue fixes this problem.

jamesarosen:
Can we add a test to make sure it happens in time for the controllers, which was the goal of 199?

jamesarosen:
(I'm not sure we can, or can easily.)

MattNguyen:
I'm not quite sure how to test this :/ any ideas?

jamesarosen:
I can't think of a better test than what we have. I'm happy as-is.

PR #208, opened 2015-05-04
Have observers set on init, component/views properties are dependent on translateable properties

dasnixon:
I noticed that when having observers set in `afterRender` queue, properties dependent on translatable properties reference an undefined translatable property since the translatable properties are not loaded until after render.

^no idea if that makes sense, queue xzbit meme

### DIFF
--- a/lib/i18n.js
+++ b/lib/i18n.js
@@ -127,7 +127,7 @@
       },
 
       _scheduleObservers: function() {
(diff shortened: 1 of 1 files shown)


jamesarosen:
See also #207 and #199

jamesarosen:
I just merged #207. I'll release a new version with that later today. If you still have trouble, let me know.

PR #211, opened 2015-05-14
v4.0 RFC

jamesarosen:
This will restart the project as an Ember-CLI addon. This commit includes the proposed API in the form of a [README](https://github.com/jamesarosen/ember-i18n/blob/v4-RFC/README.md).

I've tried to incorporate some of the best ideas from [ember-cli-i18n](https://github.com/dockyard/ember-cli-i18n). See also dockyard/ember-cli-i18n#56


jamesarosen:
@grantovich @pjmorse @jish @locks @chrmod @workmanw @bcardarella please take a look at the new README and tell me what you think. If you like this approach, I'll restart the project as an ember-cli addon and work towards the API it describes.

mohamnag:
May I have a more generic question? is this addon only intended for internationalisation or will it also provide localisation solution? I mean for things that go beyond translation like switching graphical assets, fonts, even the whole templates. That would be great to have a generic solution even though I can imagine that a service would help in that matter and by putting if blocks everywhere, which a pretty hard solution when there are many locales.

jamesarosen:
> I mean for things that go beyond translation like switching graphical assets

On projects I've worked on, we used the translation system for this. If a graphic needed to be localized, we would create a version for each locale we supported and put the URLs into the translation packages. In the structure I outlined in the RFC, that might be

```js
// app/locales/en/translations.js:

export default {
  hero_url: '/images/hero-en.png'
};
```
and used as
```hbs
<img class='hero' src="{{t 'hero_url'}}" />
```

You could do the same for fonts. Whole templates might be harder. I haven't ever come across a case where that was needed.

The other thing I've never figured out how to work into this library is number- and date formatting.

mohamnag:
that "translated urls" makes sense. However I have had cases where I really needed to replace the whole template as the translation case was getting too messy. Special cases are like when there are templates for RTL languages as well as LTR ones. This has proved in some cases really hard to put all of them in one template, but possible.

However what I find much more important is what you said about number and date formatting (I can even point to different calendars direction) probably the best solution there is to have locale aware helpers for numbers and dates so that conversion to and from values are done in those.

pjmorse:
This looks like a good approach to me. It seems like *not* being compatible with Ember-CLI is not a good long-term strategy and this setup seems to allow for broadening the functionality of the library in a way we currently can't.

I know it's early yet, but we have a Hack Day scheduled for Monday (18th) so I could throw some cycles at this then, but I don't want to force you to have all the architecture mapped out that quickly...

jamesarosen:
Some other things I'm considering for v4.0:

 * if you specify both `en-us` and `en` `translations.js` files and the user asks for `en-us`, we'll merge the two, preferring the more specific `en-us`.
 * locales can have a `settings.js` file -- e.g. `app/locales/en/settings.js`. This would be the home for things like `rtl: true`, which enables automatic insertion of RTL markers in translations. See #203.
 * `settings.js` could also be the home for `pluralForm(count)` to save an extra file
 * `settings.js` could even be a place for `formatNumber(number)`, `formatCurrency(number)`, and `formatDate(date)` if we decide to implement those at some point.

@mohamnag In theory (I'm not saying I want to implement it), we could change Ember's built-in [`resolveTemplate`](https://github.com/emberjs/ember.js/blob/v1.12.0/packages/ember-application/lib/system/resolver.js#L285-L296) to be aware of `service:i18n`. Specifically, if it's looking for `template:billing/settings` and the current `locale` is `jp`, it would first look in `app/templates/billing/settings/jp.js`, then fall back to `app/templates/billing/settings.js`. The problem with that is that there's no way to force it to rerender if `locale` changes.

The discussion of localized templates brings me to another thing I've been considering: cookbook-style documentation. For example, in order to "solve" the localized-template problem (for example), we don't build a feature into ember-i18n. Instead, we offer a series of recipes on the wiki that are "standard practice." That lets us solve problems as a community without bloating the project's official API.

README.md
```diff
@@ -2,248 +2,359 @@
...
+#### Adding Translations at Runtime
+
+If you have a translations API (so you can manage them across apps centrally
+or so you can deliver only the translations you need), you can add new
+translations at runtime via the `service:i18n`:
+
+```js
+this.i18n.addTranslations({
+  'user.profile.gravatar.help': 'Manage your avatar at gravatar.com.'
+});
```
jamesarosen: This should probably be `this.i18n.addTranslations('en', {...})`.

README.md
```diff
@@ -2,248 +2,359 @@
...
-on `Ember.I18n`, with the key and the context as arguments. You can use this to
-execute other missing-translation behaviors unrelated to the `missingMessage`,
-such as logging the key somewhere.
+on the `i18n` service, with the key and the context as arguments. You can use this
+to execute other missing-translation behavior such as logging the key somewhere.
 
-```javascript
-Ember.I18n.on('missing', function(key, context) {
+```js
+i18n.on('missing', function(key, context) {
```
jamesarosen: This should probably be `function(key, context, locale)`.

README.md
```diff
@@ -2,248 +2,359 @@
...
-customize this by overriding `missingMessage`. The below example spits out the
+customize this by overriding the function. The below example spits out the
 key along with the values of any arguments that were passed:
 
-```javascript
-Ember.I18n.missingMessage = function(key, context) {
+```js
+// app/utils/i18n/missing-translation:
+
+export default function(key, context) {
```
jamesarosen: This should probably be `function(locale, key, context, locale)`.

owilliams:
Have you considered how it could work with Pods? I'm not sure if that's still planned as the default for Ember-CLI 1.0. 

We've been splitting out our i18n files into smaller files to avoid one big translations file and keep the strings closer to the UI code.

Also, a nice to have would be nesting just the plurals. e.g.
```js
"a.long.path.to.my.string" : {
  "zero": "Nothing to see here",
  "one": "You've got 1",
  "other": "You've got loads"
}

// Instead of
"a.long.path.to.my.string.zero": "Nothing to see here",
"a.long.path.to.my.string.one": "You've got 1",
"a.long.path.to.my.string.other": "You've got loads"
```

We find the full string path more searchable in our editors than the fully nested option, but do like the non-repetitiveness of the nesting path string.

jamesarosen:
@owilliams I hadn't considered pod-based locale resolution. I've been playing around with various ways of looking up translations from the `container`, but I'm not totally happy with any of them. I really wish I could extend the application's `resolver` easily.

My guess is that breaking translations into multiple files means we'd have to iterate over all of the `requirejs` modules like [mirage does](https://github.com/samselikoff/ember-cli-mirage/blob/d04a3047c933c465fabf71f578319da30544c6f1/addon/utils/read-fixtures.js#L11-L28) to find fixtures. That *really* ties the implementation to ember-cli -- far more than relying on the `container`. That's not necessarily a deal-breaker, though.

I like the idea of breaking the keys into a long prefix and a short suffix. The current [`lookupKey`](https://github.com/jamesarosen/ember-i18n/blob/v3.1.1/lib/i18n.js#L10-L21) doesn't support that, but it does let you do

```js
{
  a: { long: { path: { to: { my: { string: {
    zero: "Nothing to see here",
    one: "You've got 1",
    other: "You've got loads"
  } } } } } }
}
```

I don't think it would be too hard to consider longer prefixes during the search. And fortunately, in the current implementation and the new one I've been playing with, that lookup is both isolated and private, so it's easy to change.

jamesarosen:
Ember-I18n v4 alpha 1 is in #213.

jamesarosen:
@owilliams this version handles
```js
export default {
  'some.long.property.name': {
    one: 'One thing',
    other: '{{count}} things'
  }
};
```
It doesn't handle breaking up translations into pods. I'm all for adding that in either v4.0 or a later minor version depending on how quickly we can flesh out the details.

acorncom:
It looks like a library from Yahoo was released back at the beginning of the year. Would it make sense for this add-on to integrate with that one in some way? Number and date formatters already seem to exist there, but this library seems more mature when handling strings ...

https://github.com/yahoo/ember-intl

blimmer:
Great work @jamesarosen , this is an excellent improvement to the project.

jamesarosen:
@acorncom I wasn't familiar with ember-intl. this rewrite is an effort to bring together ember-i18n and ember-cli-i18n. I'd be happy to talk to the Yahoo folks as well.

acorncom:
@jamesarosen Sounds good.  Just stumbled across the Yahoo extension the other day ;-)  Appreciate your hard work on this extension, we'll be making the shift from ember-cli-i18n once v4 comes out.

jasonmit:
I am all in favor of collaborating, feel free to use https://github.com/yahoo/ember-intl/issues/109 as the source of truth for this discussion.

owilliams:
Awesome work @jamesarosen, thanks!

PR #213, opened 2015-05-18
Restart as Ember-CLI addon

jamesarosen:
This will restart the project as an Ember-CLI addon.

https://github.com/dockyard/ember-cli-i18n/issues/56 is worthwhile background reading for this PR.

 * Restart project as a native Ember-CLI addon
 * Translations now recompute when locale changes
 * Move central API to `service:i18n`, which exposes `locale`, `t`, and `addTranslations`
 * Add `translationMacro` for defining translated computed properties
 * Remove `eachTranslatedAttribute`, `TranslateableProperties`, and `TranslateableAttributes`, which are better expressed as computed property macros
 * Built-in compiler now applies RTL Unicode markers *before* marking the result HTML-safe. (Previously, `rtl = true` broke HTML safety.)

**HUGE** thanks to everyone who has contribute to this project and to https://github.com/dockyard/ember-cli-i18n, which strongly informed this rewrite.


jamesarosen:
This won't work with Glimmer until https://github.com/emberjs/ember.js/issues/11080 is resolved.

sandstrom:
Nice! :boat:

.ember-cli
```diff
@@ -0,0 +1,9 @@
+{
+  /**
+    Ember CLI sends analytics information by default. The data is completely
+    anonymous, but there are times when you might want to disable this behavior.
+
+    Setting `disableAnalytics` to true will prevent any data from being sent.
+  */
+  "disableAnalytics": false
+}
```
pjmorse: Do we need this? Or is it more important that this file exist, even if the content is close to a no-op?

jamesarosen: We don't, actually. Here's how ember-cli calculates `disableAnalytics`: https://github.com/ember-cli/ember-cli/blob/b915143cd2e4d2067cebcd4dc93c2bf4d5b98afa/lib/cli/index.js#L57-L61. And here's how Yam parses files: https://github.com/twokul/yam/blob/master/lib/config.js#L10-L25. So if the file is missing, Yam treats it as an empty Object, which means `config.get('disableAnalytics')` will be just as falsy as it is now.

listepo:
:+1:

PR #216, opened 2015-05-19
Ember 2.0 attrs support

miguelmota:


bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ember-i18n",
-  "version": "3.1.1",
+  "version": "3.1.2",
```
jamesarosen: Would you mind resubmitting this without the version bump?

lib/i18n.js
```diff
@@ -126,18 +130,18 @@
         });
       },
 
-      _scheduleObservers: Ember.on('init', function() {
+      _scheduleObservers: function() {
```
jamesarosen: Not all apps have prototype extensions enabled. The official policy is that addons *shouldn't* rely on them.

locks: The official policy is that addons have prototypes turned off by default actually (starting with 0.2.x?), so this is good advice

locks:
@miguelmota as a rule of thumb you shouldn't bump versions in PRs, since the maintainer of the library usually wants to control the versioning.

The 1.10 branch is failing the tests because the HTMLBars compiler isn't required it seems.

jamesarosen:
v4.0 will support Ember 2.0. The v3.x branch never will.

PR #227, opened 2015-06-08
Ember 1.13 and 2.0 support

jamesarosen:
Starting work on support for Ember 1.13. Glimmer breaks the old helper and adds a new `Ember.Helper` with more of the power we need.

As of this commit, the beta build is still breaking because there's something wrong with the helper registration.


jamesarosen:
@mixonic thanks for your work on `Ember.Helper`! I haven't quite got it working in ember-i18n, but I feel like we're closing in :)

app/instance-initializers/ember-i18n.js
```diff
@@ -13,7 +14,14 @@ export default {
     }
     instance.container.lookup('service:i18n').set('locale', defaultLocale);
 
-    Ember.HTMLBars._registerHelper('t', helper);
+    if (legacyHelper != null) {
+      Ember.HTMLBars._registerHelper('t', legacyHelper);
+    }
+
+    if (Helper != null) {
+      instance.registry.register('helper:t', Helper);
```
jamesarosen: Despite this, I get a "helper t not found" error message whenever I try to use it.

mixonic:
Awesome. The dashless stuff should land soon as well. /cc @rwjblue

app/instance-initializers/ember-i18n.js
mixonic: Correct. Robert is going to land https://github.com/emberjs/rfcs/pull/58 into 1.13 which will give you `{{t` without any explicit registration step.

jamesarosen: Ah. I was basing this off of https://github.com/emberjs/ember.js/pull/11278/files#diff-5d4704d516d9bedb9605e499dc0a88d8R122

I guess once dashless-autolookup lands I'll want a single `app/helpers/t.js` that exports either `legacyHelper` or `Helper`, depending on the presence of `Ember.Helper`. That'll be easy :)

rwjblue: Confirm.

jamesarosen:
Though this is meant to work with 1.13 (currently beta), there isn't a 1.13 beta that includes emberjs/ember.js#11278 and emberjs/ember.js#11393, so the best way to work on this is against canary.

The current problem I'm having is that `Helper#recompute` fires, but for some reason, it doesn't subsequently call `#compute`.

Some relevant links:

 * [the test](https://github.com/jamesarosen/ember-i18n/blob/master/tests/acceptance/t-helper-test.js#L46-L50)
 * [an important test helper](https://github.com/jamesarosen/ember-i18n/blob/master/tests/helpers/start-app.js#L21-L28)

jamesarosen:
After much debugging help from @mixonic, it seems to be a problem with using `recompute` and an ordered param. In his words:

> adding an argument means linkedParams are taken, and our later linkParams call fails

mixonic:
@jamesarosen https://github.com/emberjs/ember.js/pull/11445 contains the fix. A merge today seems likely, and a 1.13.1 on Tuesday.

jamesarosen:
The last test will be fixed by emberjs/ember.js#11445

rwjblue:
https://github.com/emberjs/ember.js/pull/11445 has been merged, and is in both beta and release channels.

Could you test, and confirm the release channel fixes the remaining failures here?

josepjaume:
Just a heads up: I think this branch breaks the bound `{{t}}` helper; when I change `i18n.locale`, the translations don't update.

jamesarosen:
@josepjaume have you tested with a version of Ember since emberjs/ember.js#11445 was merged?

@rwjblue tests look great! Thanks :)

josepjaume:
Oh, I see! Does this mean there's a `1.13.1` bugfix release coming soon, or is this behavior broken until `2.0.0`?

josepjaume:
BTW: Thanks for the good work :+1:

jamesarosen:
> Does this mean there's a 1.13.1 bugfix release coming soon?

I'm [told](https://github.com/jamesarosen/ember-i18n/pull/227#issuecomment-111876471) it'll be Tuesday.

josepjaume:
Confirmed it's working by using the `release` channel. In my `bower.json` file:

```json
  "dependencies": {
    "ember": "components/ember#release",
    [...]
  },
  "resolutions": {
    "ember": "release"
  }
}
```

Eptis:
Don't know if this is still supported, but this subexpression no longer works:
```
{{moment date.date (t "app.defaultMomentFormat")}}
```

PR #229, opened 2015-06-11
Explain that you can't translate until `ember-i18n` initializer runs

cibernox:
This bite me yesterday. 

Since the order of the initializers is not enforced, some could do translations and other couldn't because the locale was not defined yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,8 +102,8 @@ export default {
   name: 'my-initializer',
   after: 'ember-i18n',
   initialize: function({ container }) {
-    let i18n = container.lookup('service:i18n');
-    i18n.t('some.translation');
+    const i18n = container.lookup('service:i18n');
+    document.title = i18n.t('some.translation');
   }
 };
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -92,6 +92,21 @@ export default Ember.Object.extend({
...
+
+```js
+// app/instance-initializers/my-initializer.js
+
+export default {
+  name: 'my-initializer',
+  after: 'ember-i18n',
+  initialize: function({ container }) {
+    let i18n = container.lookup('service:i18n');
+    i18n.t('some.translation');
```
jamesarosen: I like this documentation block, but it might help to flesh out the use-case just a tiny bit:
```js
const i18n = container.lookup('service:i18n');
document.title = i18n.t('document.title');
```

jamesarosen:
Generally for this.

cibernox:
Updated example

jamesarosen:
:+1: :)

PR #230, opened 2015-06-11
[Feature] Allow to pass an array of fallback translations

cibernox:
I'm need to create a helper that translates error messages following the same conventions that rails-i18n uses for translate error messages in active record.

For those not familiar with rails-i18n, the fallback chain that it uses is like this (being `Admin` a subclass of `User`)
```
activerecord.errors.models.admin.attributes.name.blank
activerecord.errors.models.admin.blank
activerecord.errors.models.user.attributes.name.blank
activerecord.errors.models.user.blank
activerecord.errors.messages.blank
errors.attributes.name.blank
errors.messages.blank
```

While this is specific of my in progress meta-addon (addon built on top of ember-i18n :boom: ), the ability of having fallback translations is a must for me and I think general purpose feature. 

ATM i need to compare if the returned string matches the regular expression `/Missing translation/` or alternatively use the private API `i18n._locale`.

The downside of this change is that `fallbacks` becomes keyword-ish. I'm open to other ideas about how to implement this.


jamesarosen:
> ATM i need to compare if the returned string matches the regular expression /Missing translation/ or alternatively use the private API i18n._locale.

We [mark](https://github.com/jamesarosen/ember-i18n/blob/v4.0.0-beta.2/addon/locale.js#L68) missing translations with `_isMissing = true`. The `service:i18n` doesn't expose a way to get the compiled translation at the moment, but it could.

addon/service.js
```diff
@@ -19,8 +19,20 @@ export default Ember.Object.extend(Ember.Evented, {
   t: function(key, data = {}) {
     const locale = this.get('_locale');
     Ember.assert("I18n: Cannot translate when locale is null", locale);
+    const fallbacks = get(data, 'fallbacks');
```
jamesarosen: The logic below will be a little cleaner if you do
```js
const fallbacks = getWithDefault(data, 'fallbacks', []);
```

addon/service.js
```diff
@@ -19,8 +19,20 @@ export default Ember.Object.extend(Ember.Evented, {
...
+    const count = get(data, 'count');
 
-    const template = locale.getCompiledTemplate(key, get(data, 'count'));
+    let template = locale.getCompiledTemplate(key, count);
+
+    if (template._isMissing && fallbacks) {
+      for (let i = 0; i < fallbacks.length; i++) {
```
jamesarosen: If you unshift `key` onto `fallbacks`, you can treat it just like a fallback:
```js
fallbacks.unshift(key);
const template = fallbacks.find(function(key) {
  const result = locale.getCompiledTemplate(key, count);
  return !result._isMissing;
});
```

jamesarosen: Oh, except that might end up with `undefined`, so you'd then need to pick the last fallback. That's not really any nicer :/

jamesarosen: Well, the `unshift` does help, but only with the `for` version:
```js
const keys = getWithDefault(data, 'fallbacks', []);
keys.unshift(key);
let template;
for (let i = 0; i < keys.length; i++) {
  template = locale.getCompiledTemplate(keys[i], count);
  if (!template._isMissing) { break; }
}
```

jamesarosen: It *might* make sense to pass all the keys into `getCompiledTemplate` and have the `Locale` be responsible for fallback behavior.

cibernox:
@jamesarosen Mixin the main key with the fallbacks in a single array and let `Locale#getCompiledTemplate` do it's job made a lot of sense.

The point am not 100% convinced yet is the fallbacks being specified as a `fallbacks` option in the hash. Maybe it makes more sense to allow the `key` itself to be an array. What do you think?

cibernox:
Updated this PR to almost mimic 1:1 rails-i18n behaviour (see http://guides.rubyonrails.org/i18n.html#defaults)

The array of fallbacks is received in the options hash and the name of the key with the fallback translations is named `default`. 
Like rails, that key can contain a single default value or an array of them that are applied in order.

Examples:
```
// Transltions: { "puppies.cat": "kitty", "puppies.dog": "cub" } 
i18n.t('puppies.tiger', { default: 'puppies.cat' }) // => "kitty"
i18n.t('puppies.wolf', { default: ['puppies.fox', 'puppies.dog'] }) // => "cub"
i18n.t('puppies.ferret', { default: ['puppies.polecat'] }) // => "Translation Missing: puppet.ferret"
```

cibernox:
Tests are failing because of the glimmer incompatibility.

jamesarosen:
> Tests are failing because of the glimmer incompatibility.

You can either rebase off master now that the beta and canary tests are marked as *may-fail*, or wait until emberjs/ember.js#11445 and then #227 are merged.

cibernox:
Green :grin:

cibernox:
Now that 4.0 is out of the door, can we revisit this?

addon/locale.js
```diff
@@ -32,29 +32,36 @@ export default class Locale {
...
+      }
     }
 
     if (result == null) {
-      result = this._defineMissingTranslationTemplate(key);
+      result = this._defineMissingTranslationTemplate(fallbackChain[0]);
     }
 
-    Ember.assert(`Template for ${key} in ${this.id} is not a function`, Ember.typeOf(result) === 'function');
+    Ember.assert(`Template for ${fallbackChain[0]} in ${this.id} is not a function`, Ember.typeOf(result) === 'function');
```
jamesarosen: This might be confusing, since it could be one of the fallbacks that's wrong. I'm wondering whether we should save off the key of the translation we found so we can use it here.

addon/service.js
```diff
@@ -20,8 +20,15 @@ export default Parent.extend(Ember.Evented, {
   t: function(key, data = {}) {
     const locale = this.get('_locale');
     Ember.assert("I18n: Cannot translate when locale is null", locale);
+    const count = get(data, 'count');
 
-    const template = locale.getCompiledTemplate(key, get(data, 'count'));
+    let defaults = get(data, 'default') || [];
+    if (!isArray(defaults)) {
+      defaults = [defaults];
+    }
```
jamesarosen: This could be `Ember.makeArray(get(data, 'default'))`.
```js
Ember.makeArray(undefined); // []
Ember.makeArray(null); // []
Ember.makeArray('foo'); // [ 'foo' ]
Ember.makeArray([ 'foo', 'bar' ]); // [ 'foo', 'bar' ]
```

tests/unit/i18n-t-test.js
```diff
@@ -54,3 +54,12 @@ test("applies custom pluralization rules", function(assert) {
...
 });
+
+test("applies provided default translation in cascade when main one is not found", function(assert) {
+  const i18n = this.subject({ locale: 'en' });
+  const defaultsChain = ['with.pretty-good-interpolations', 'with.interpolations'];
+
+  assert.equal(i18n.t('with.great-interpolations', { clicks: 8, default: defaultsChain }), 'Clicks: 8', 'default can be an array');
+  assert.equal(i18n.t('with.great-interpolations', { clicks: 8, default: 'with.interpolations' }), 'Clicks: 8', 'default can be an string');
+  assert.equal(i18n.t('not.yet.translated', { clicks: 8, default: ['not.translated.either'] }), 'Missing translation: not.yet.translated');
+});
```
jamesarosen: I'd also like to see a test that the missing event happens with the *primary* key.

jamesarosen:
I'm 100% sold on the idea and the API! I made a couple comments about smaller details.

addon/service.js
cibernox: Nice trick. Like `Kernel.Array()` in ruby.

addon/locale.js
cibernox: Done

tests/unit/i18n-t-test.js
cibernox: The line 64 tests that, isn't it?

jamesarosen: Sort of. It tests that the primary key gets sent to the missing-translation helper. The event is different, but I'm not sure the distinction is important.

cibernox: Oh, i get you know. I wasn't even aware that there was an event. I'll test that too.

cibernox:
Feedback applied.

addon/locale.js
```diff
@@ -32,29 +32,36 @@ export default class Locale {
...
+      }
     }
 
     if (result == null) {
-      result = this._defineMissingTranslationTemplate(key);
+      result = this._defineMissingTranslationTemplate(fallbackChain[0]);
     }
 
-    Ember.assert(`Template for ${key} in ${this.id} is not a function`, Ember.typeOf(result) === 'function');
+    Ember.assert(`Template for ${fallbackChain[i]} in ${this.id} is not a function`, Ember.typeOf(result) === 'function');
```
jamesarosen: This should be `0`, not `i`, right?

addon/service.js
```diff
@@ -25,8 +25,12 @@ export default Parent.extend(Ember.Evented, {
   t: function(key, data = {}) {
     const locale = this.get('_locale');
     Ember.assert("I18n: Cannot translate when locale is null", locale);
+    const count = get(data, 'count');
 
-    const template = locale.getCompiledTemplate(key, get(data, 'count'));
+    let defaults = makeArray(get(data, 'default'));
```
jamesarosen: This could be `const`, but `let` is just fine. If we ever come back to this function, we can change it then.

addon/service.js
```diff
@@ -25,8 +25,12 @@ export default Parent.extend(Ember.Evented, {
...
+
+    defaults.unshift(key);
+    let template = locale.getCompiledTemplate(defaults, count);
```
jamesarosen: ibid

jamesarosen:
+1 merging and squashing...

jamesarosen:
Closed in 8aa758cb1c4dd6f05442ddafef3f6f2e1609141b

cibernox:
wohoo!

jamesarosen:
I'd like to document this in the wiki, but I'm not sure of a good spot. Maybe at the bottom of [Translating Text](https://github.com/jamesarosen/ember-i18n/wiki/Doc:-Translating-Text) in a new "Fallback Translation Keys" section?

cibernox:
Yes, that seems a good place. That section basically teaches all the usages of the library.

PR #231, opened 2015-06-11
Fix spanish grammar

cibernox:
Remove wrong accent :+1:


jamesarosen:
I wonder whether that's regional. I checked it with a Mexican-Spanish speaker. From my knowledge of Spanish, it does seem odd to put an accent where the natural stress would be, as I've done.

+1

cibernox:
In this case it's not regional. The singular one (interpolación) has accent but not the plural.

jamesarosen:
TIL!

That follows my rule of *add an accent when the emphasis doesn't fall in the natural place*, since the natural emphasis for the singular would be *interpolacEEon*.

PR #232, opened 2015-06-11
[BUG] Dynamic keys are not being translated

cibernox:
Things like `{{t dynamicKey}}` are not working ATM. This PR includes a
failing test.

The {{t}} helper (signature t(params, hash, options, env)) is receiving
a an array if `Stream` objects in `params`, not the actual values coming
from that stream.

I suspect (because I saw this in the past) that is because of the way
the helper is being registered. The registration of helpers has changed
a few times since HTMLBars.

### DIFF
--- a/tests/acceptance/t-helper-test.js
+++ b/tests/acceptance/t-helper-test.js
@@ -23,6 +23,18 @@ test('renders with static interpolations', function(assert) {
   assert.textIs('.static-interpolations', 'Clicks: 45');
 });
 
+test('renders translations of dynamic keys', function(assert) {
+  assert.textIs('.dynamic-key', 'text with no interpolations');
+});
+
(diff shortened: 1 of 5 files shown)


jamesarosen:
Are you sure? This test passes for me. And [the helper does seem to handle that case](https://github.com/jamesarosen/ember-i18n/blob/v4.0.0-beta.2/addon/helper.js#L26-L29).

jamesarosen:
Oh, sorry. The tests are failing. The problem is not that it doesn't *observe* updates, but that it doesn't *evaluate* streams. Working on that!

jamesarosen:
In your test, the route was setting `dynamicKey` on itself, not the `controller`. I've fixed that as well as the implementation and pushed as f2c31684759931f2450ab111139586a1c6472b7c.

PR #235, opened 2015-06-12
Don't auto-inject i18n into routes, controllers, components

### LINKED ISSUE #226: Don't inject the i18n service automatically
I don't think the i18n service needs to be blindly injected into all `Route`, `Controller`, and `Component`. It's a lazy lookup, but I still think it's easier to just do:
```js
i18n: Ember.inject.service();
```
where you need i18n.

This is a style preference, not really a direct issue.

### LINKED ISSUE #234: Bad Docs? this.i18n.t() does not exist but this.get('i18n').t() does
Inside a route i'm doing 

`i18n: Ember.inject.service();`
`this.i18n.t('login.loginTitle');`

But the function does not exist when called this way.
But when i call this way it is working. 

`i18n: Ember.inject.service();`
`this.get('i18n').t('login.loginTitle');`

Are the docs just wrong or is this something different?

jamesarosen:
Instead, recommend `i18n: Ember.inject.service()` or declaring your own injections in an instance-initializer.

Closes #226
Closes #234
See also #229

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,29 +67,35 @@ $ ember generate translations es
 
 ### `i18n` Service
 
-Many pieces of ember-i18n rely on the `service:i18n`. This service is automatically
-injected into every Route, Controller and Component in your app. If you need it
-elsewhere, you can register your own injection:
+Many pieces of ember-i18n rely on the `service:i18n`. Ask for it wherever you
+need (likely in `Route`s and `Component`s):
 
 ```js
-// app/initializers/i18n.js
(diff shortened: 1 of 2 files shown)


jamesarosen:
@blimmer what do you think?

blimmer:
@jamesarosen this is way more in line with what I'd expect. Also, people can always inject the service everywhere if they prefer that style. It's better for the user to make that decision vs. the addon in my mind.

PR #242, opened 2015-06-18
support podModulePrefix

### LINKED ISSUE #241: Doesn't work with podModulePrefix set
If I add `podModulePrefix: 'dummy/pods'` to the app's `config/environment.js`, the translations suddenly all disappear.

The problem is the `lookupFactory` call:

```js
// podModulePrefix: undefined
container.lookupFactory('translations:locales/en'); // { ... }
container.lookupFactory('locale:en/translations'); // undefined

// podModulePrefix: 'dummy/pods'
container.lookupFactory('translations:locales/en'); // undefined
container.lookupFactory('locale:en/translations'); // { ... }
```

I don't know of a way to support both.

See also https://github.com/ember-cli/ember-cli/issues/3424

/cc @rwjblue @sethphillips

jamesarosen:
Previously, we were looking up `translations:locales/en`. That works with `podModulePrefix` unset, but breaks with it set. Instead, we turn the lookup string around: `locale:en/translations`.

This adds an additional test run with `podModulePrefix` set to `dummy/pods`.

Resolves #241

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ cache:
 
 env:
   - EMBER_TRY_SCENARIO=default
+  - EMBER_TRY_SCENARIO=default POD_MODULE_PREFIX="dummy/pods"
   - EMBER_TRY_SCENARIO=ember-release
   - EMBER_TRY_SCENARIO=ember-beta
   - EMBER_TRY_SCENARIO=ember-canary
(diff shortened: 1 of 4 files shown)


jamesarosen:
@sethphillips can you try this out on your app?

```json
// package.json
"ember-i18n": "jamesarosen/ember-i18n#pod-module-prefix"
```

Kilowhisky:
@jamesarosen I'm using the podModulePrefix and just tried it out and all is well. It was able to find the locale files correctly.

sethphillips:
That did the trick my friend!  Well done tracking that down.

PR #245, opened 2015-06-19
Swap injection info to initializer in README.

rwjblue:
My simple rule of thumb for when to use an initializer vs an instance initializer is: 

* when I need to setup things for lookup at a later time, use an initializer
* when I need to look something up and perform operations on it, use an instance initializer

*tldr;* All factories and injection rules should be done before instance initializers are ran (so that they can look things up properly).


jamesarosen:
+1

Issue #247, opened 2015-06-21
listing of installed locales


kellyselden:
I'm porting my projects from ember-cli-i18n. I'm about to port over my addon [ember-cli-i18n-locales](https://github.com/kellyselden/ember-cli-i18n-locales) which exposes a util function that returns the locale names installed for use in a locale chooser component. Any interest in a PR, or would you prefer to keep it out in an addon?

jamesarosen:
I don't think a locale-chooser component is right, but adding a single public CP to `service:i18n` that lists the locales in the registry seems useful.

kellyselden:
Right, the locale chooser component would be implemented in an app, not this addon.

jamesarosen:
If you wanted, you could also emit a warning when changing `locale`:
```js
locale: Ember.computed({
  set(newLocale) {
    Ember.warn('No translations registered for ' + newLocale, function() {
      return this.get('availableLocales').some(function(loc) {
        return loc.indexOf(newLocale) === 0;
      });
    });

    return newLocale;
  }
})
```

kellyselden:
fixed in #250

PR #250, opened 2015-06-21
[Feature] add locales CP to the service

kellyselden:
A result of #247. Adds a way to get a list of the added locales. Useful for rendering this list to a template to switch locales.

### DIFF
--- a/addon/get-locales.js
+++ b/addon/get-locales.js
@@ -1,0 +1,13 @@
+import Ember from 'ember';
+
+const matchKey = '/locales/(.+)/translations$';
+
+export default function getLocales() {
+  return Ember.keys(requirejs.entries).reduce((locales, module) => {
+    var match = module.match(matchKey);
+    if (match) {
+      locales.pushObject(match[1]);
+    }
+    return locales;
+  }, Ember.A());
+}
--- a/tests/unit/locales-test.js
+++ b/tests/unit/locales-test.js
@@ -1,0 +1,20 @@
+import Ember from 'ember';
+import { moduleFor, test } from 'ember-qunit';
+
+moduleFor('service:i18n', 'I18nService#locales', {
+  integration: true
+});
+
+const locales = [
+  'en-bw',
+  'en-ps',
+  'en-wz',
+  'en',
+  'es'
+];
+
+test('locales have the correct length and names', function(assert) {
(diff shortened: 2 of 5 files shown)


addon/get-locales.js
```diff
@@ -0,0 +1,13 @@
...
+
+const matchKey = '/locales/(.+)/translations$';
+
+export default function getLocales() {
+  return Ember.keys(requirejs.entries).reduce((locales, module) => {
+    var match = module.match(matchKey);
+    if (match) {
+      locales.push(match[1]);
+    }
+    return locales;
```
jamesarosen: This should probably wrap the result in `Ember.A()`. I think that's the best practice for things that might end up in a Glimmer `each`.

tests/unit/locales-test.js
```diff
@@ -0,0 +1,30 @@
...
+
+  assert.equal(i18n.get('locales').length, locales.length);
+});
+
+test('locales have the correct names', function(assert) {
+  const i18n = this.subject();
+
+  assert.expect(locales.length);
+
+  locales.forEach(locale => {
```
jamesarosen: You could use `assert.deepEqual` here, though you might have to `sort` the results.

addon/get-locales.js
kellyselden: I went back and forth on that. I will add it.

tests/unit/locales-test.js
kellyselden: Good call. Had to reorder the expected.

kellyselden:
fixed the comments.

addon/get-locales.js
```diff
@@ -0,0 +1,13 @@
...
+const matchKey = '/locales/(.+)/translations$';
+
+export default function getLocales() {
+  return Ember.keys(requirejs.entries).reduce((locales, module) => {
+    var match = module.match(matchKey);
+    if (match) {
+      locales.pushObject(match[1]);
+    }
+    return locales;
+  }, Ember.A());
```
jamesarosen: Nice use of `reduce`!

jamesarosen:
:octopus: :moneybag: :tangerine:

addon/get-locales.js
kellyselden: Thanks!

PR #258, opened 2015-06-25
Destroy stream once view is destroyed

danshapir:


jamesarosen:
Can we add some sort of test? Maybe something like
```hbs
{{#if showSubexpressionExample}}
  {{input name="email" placeholder=(t "email.placeholder")}}
{{/if}}
```
and then toggle `showSubexpressionExample`. That way, we'll destroy the view that is using the helper.

danshapir:
Sure, but to be honest, never used the testing with ember, so I've got no idea how to do it.

jamesarosen:
I can't get this failure condition to show up in tests at all. We'll just have to scrap the tests :/

PR #262, opened 2015-06-30
Auto fallback to the en config

danshapir:
As a big company using this addon we often add new languages without deploying the application (Mobile App), so we don't have all the locales config files on the users end, we would like the addon to allow a fallback to en if no other config was found.

P.S
Added a test and it's working great

### DIFF
--- a/addon/locale.js
+++ b/addon/locale.js
@@ -20,15 +20,25 @@ export default class Locale {
 
   rebuild() {
     this.translations = getFlattenedTranslations(this.id, this.container);
-
+    this._setConfig();
+  }
+  
+  _setConfig() {
     walkConfigs(this.id, this.container, (config) => {
       if (this.rtl === undefined) { this.rtl = config.rtl; }
       if (this.pluralForm === undefined) { this.pluralForm = config.pluralForm; }
     });
-
-    if (this.rtl === undefined) { this.rtl = false; }
+  
+    const defaultConfig = this.container.lookupFactory('ember-i18n@config:zh');
+  
+    if (this.rtl === undefined) {
+      Ember.warn('ember-i18n: No RTL configuration found for ${this.id}.');
+      this.rtl = defaultConfig.rtl;
+    }
+  
     if (this.pluralForm === undefined) {
-      throw new Error(`No pluralForm found for ${this.id}`);
+      Ember.warn('ember-i18n: No pluralForm configuration found for ${this.id}.');
+      this.pluralForm = defaultConfig.pluralForm;
     }
   }
 
@@ -109,10 +119,6 @@ function walkConfigs(id, container, fn) {
 
   const parentId = parentLocale(id);
   if (parentId) { walkConfigs(parentId, container, fn); }
-  
-  // Fallback to en at the end, so that even if the locale is not conigured no failure will happen
-  const enConfig = container.lookupFactory(`ember-i18n@config:en`);
-  if (enConfig) { fn(enConfig); }
 }
 
 function parentLocale(id) {
--- a/addon/locale.js
+++ b/addon/locale.js
@@ -109,6 +109,10 @@ function walkConfigs(id, container, fn) {
 
   const parentId = parentLocale(id);
   if (parentId) { walkConfigs(parentId, container, fn); }
+  
+  // Fallback to en at the end, so that even if the locale is not conigured no failure will happen
+  const enConfig = container.lookupFactory(`ember-i18n@config:en`);
+  if (enConfig) { fn(enConfig); }
 }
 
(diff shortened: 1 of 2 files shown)


kellyselden:
There is no sense in this library that 'en' is the default language. Hardcoding 'en' is not a good idea.

danshapir:
Do you prefer falling back to the default language's config?

jamesarosen:
Ah, you're not asking that English *translations* be served by default, just that the config (RTL and pluralization rules) be default. I think we should have a default, but that it should be the same as Chinese: LTR and no pluralization rules at all.

jamesarosen:
Actually, I think the default should be whatever the app sets as `defaultLocale` and that the fallback mechanism use that as its root.

pjmorse:
I agree with @jamesarosen's last comment: rather than express (code) an opinion, better to let the app designate a default.

kellyselden:
I recant my earlier comment. I see now that 'en' is the fallback default.

danshapir:
As I said in my previous comment, default locale shouldn't be the configuration fallback. If your app has default locale arabic, adding a new ltr locale shouldn't fallback to Arabic. Maybe adding another configuration, defaultConfig is an option. But for some reason I couldn't access the environment for from the locale.js file.

jamesarosen:
@danshapir why are you letting your users select a locale that your app doesn't support? And what do you really expect the app to do in that case? I'm beginning to think this isn't a problem that the library should solve. That is, your app supports locales `ab` and `cd`. You *must* pick one of those to make the app work. How you pick is up to you.

danshapir:
@jamesarosen Since this addon supports adding locales on the fly, we are loading the translations from a WebAPI and the JS code is a PhoneGap/Cordova mobile application, meaning not updated by a simple refresh (though that's not a reasonable solution as well). If you've shipped it with support for lets say EN, GR and FR, and now your API has added a new locale, there is no reason to ship the product again since the only drawback is the config file that is throwing an exception (plurarForm) if you select a language that doesn't already contain a locale config. I think that since the plurarForm might not be used by all customers (I know over 15 major application using this addon that don't manage the plurar through it) you shouldn't be killed by a missing config, a fallback should always be made.

Also as a concept, an addon shouldn't throw an error that kills the application for such a thing as switching to a locale. The end used has added the translations - unless you want to support injecting config in run time, but that seems weird.

jamesarosen:
It sounds like the core problem is that ember-i18n lets you define translations for a locale it doesn't know about, but not config. Thus, we have a few options:

 1. add `i18n.defineConfig(locale, config)`. Would that *replace* any existing config for that locale or be merged into it? I suspect the latter, as that is how `addTranslations` works.
 2. put an arbitrary config at the root of the `walkConfigs` chain. Any choice would be solipsistic, but the simplest would probably be Chinese, which is LTR and whose pluralization rule is simply `return OTHER;` -- both of those are very common among human languages.

danshapir:
I would suggest both, the first because it seems logical to allow that feature since we allow adding translations on the fly.
The second since we should never kill the user's application.

jamesarosen:
`Locale` already [sets a default for `rtl`](https://github.com/jamesarosen/ember-i18n/blob/v4.0.0/addon/locale.js#L29). Instead of [throwing an error for missing `pluralForm`](https://github.com/jamesarosen/ember-i18n/blob/v4.0.0/addon/locale.js#L31), I suggest we emit a warning and then use the value from Chinese. In fact, I recommend warning for `rtl` missing as well. Thus, something like

```js
rebuild() {
  this.translations = getFlattenedTranslations(this.id, this.container);
  this._setConfig();
},

_setConfig() {
  walkConfigs(this.id, this.container, (config) => {
    if (this.rtl === undefined) { this.rtl = config.rtl; }
    if (this.pluralForm === undefined) { this.pluralForm = config.pluralForm; }
  });

  const defaultConfig = container.lookupFactory(`locale:zh/config`);

  if (this.rtl === undefined) {
    warn(`ember-i18n: No RTL configuration found for ${this.id}.`);
    this.rtl = defaultConfig.rtl;
  }

  if (this.pluralForm === undefined) {
    warn(`ember-i18n: No pluralForm configuration found for ${this.id}.`);
    this.pluralForm = defaultConfig.pluralForm;
  }
}
```

danshapir:
Sounds good to me! Just allow the user to turn those warnings of in production mode.

jamesarosen:
`Ember.warn` should do that automatically

danshapir:
I'll update the pr tomorrow!

danshapir:
https://github.com/jamesarosen/ember-i18n/pull/263

Updated

PR #263, opened 2015-07-02
Auto fallback to the zh config for both plural & rtl

danshapir:


tests/unit/i18n-t-test.js
```diff
@@ -67,3 +67,9 @@ test("applies provided default translation in cascade when main one is not found
   assert.equal(i18n.t('not.yet.translated', { clicks: 8, default: ['not.translated.either'] }), 'Missing translation: not.yet.translated');
   assert.equal(calls[0][1], 'not.yet.translated', 'When the "missing" event is fired, it\'s fired with the provided key');
 });
+
+
+test("check unknown locale", function(assert) {
```
kellyselden: this isn't testing the rtl or pluralForm changes.

danshapir:
It is testing that a fallback was made, I don't think the rest should check which since it doesn't matter. The main idea is to pass the test is an unknown locale.  But you are more than welcome to add tests if you want...

kellyselden:
You're testing that it doesn't throw an error anymore. If you were to comment out `this.rtl = defaultConfig.rtl;` or `this.pluralForm = defaultConfig.pluralForm;`, your test would still pass I think, so it has potential to break in the future.

danshapir:
That's not the meaning of this pr, the default config as a concept is... Not whether it's rtl or not, but again, you are more than welcome to contribute!

jamesarosen:
If the documentation doesn't specify what the default `rtl` and `pluralForm` are, then I don't think we need specific tests for them.

*However*, the test for this should at least send in `{ count: anyNumber }` to ensure that `pluralForm` _is either ignored or works_. (In this implementation, it works, but ignoring it by default would be a legal implementation given that test. What wouldn't be is trying to call a function that doesn't exist.)

danshapir:
Done

jamesarosen:
I dig it.

PR #270, opened 2015-07-08
Improve config-missing warnings

jamesarosen:
The new `Locale#_setConfig` introduced in #263 relies on *some* config in the path defining `rtl`. If it's `undefined`, `_setConfig` now emits a warning. Previously, we were relying on `undefined` being falsy, but we need to specify a value (`false`) to prevent the warning. We define `rtl: false` for en, fr, hi, ru, and zh. All other included locales either define `rtl: true` or inherit from one of these.

Additionally, the warnings added in #263 used ES6 string interpolation, but didn't have the correct ES6 string syntax. This corrects that, which makes for better error messages.

### DIFF
--- a/addon/config/en.js
+++ b/addon/config/en.js
@@ -1,6 +1,8 @@
 import { ONE, OTHER } from "./constants";
 
 export default {
+  rtl: false,
+
   pluralForm: function(n) {
     if (n === 1) { return ONE; }
     return OTHER;
--- a/addon/config/fr.js
+++ b/addon/config/fr.js
@@ -1,6 +1,8 @@
 import { ONE, OTHER } from "./constants";
 
 export default {
+  rtl: false,
+
   pluralForm: function(n) {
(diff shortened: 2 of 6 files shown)


jamesarosen:
@kulja can you try this out and let me know if it works?
```json
// package.json
"ember-i18n": "jamesarosen/ember-i18n#rtl-false"
```
and then
```bash
rm -rf node_modules/ember-i18n
npm i
```

Kulja:
Warning is gone for locale `en` (I don't have config.js for this locale).
Locale `hr` for which I have config.js `rtl: false' is still needed (I am guessing that is ok?)

jamesarosen:
ember-i18n doesn't pre-define `hr`, so you'll need to do that:
```js
// app/locales/hr/config.js:
export default {
  rtl: false,
  pluralForm: ...
};
```

Kulja:
Great, then it is all fine.

PR #273, opened 2015-07-14
add test helpers

iezer:
@jamesarosen Adds some test helpers that can be imported into any app that uses the add-on.

To use in an app,  in `tests/helpers/start-app.js` add this line: 

`import './ember-i18n/test-helpers';`

You can use the `t` helper like this:  ``find(`.header:contains(${t('welcome_message')})`)``

and the `expectTranslation` helper like this: `expectTranslation('.header', 'welcome_message');`

[blog post](http://williamsbdev.com/posts/Ember-CLI-test-support/) about including test helpers in the `test-support` directory.

After this is merged I'll add some documentation to the wiki.

### DIFF
--- a/test-support/helpers/ember-i18n/test-helpers.js
+++ b/test-support/helpers/ember-i18n/test-helpers.js
@@ -1,0 +1,31 @@
+import Ember from 'ember';
+
+// example usage: find(`.header:contains(${t('welcome_message')})`)
+Ember.Test.registerHelper('t', function(app, key, interpolations){
+  const i18n = app.__container__.lookup('service:i18n');
+  return i18n.t(key, interpolations);
+});
+
+// example usage: expectTranslation('.header', 'welcome_message');
+Ember.Test.registerHelper('expectTranslation', function(app, element, key, interpolations){
+  const text = app.testHelpers.t(key, interpolations);
+
+  assertTranslation(element, key, text);
+});
+
+const assertTranslation = (function() {
+  if (typeof QUnit !== 'undefined' && typeof ok === 'function') {
+    return function(element, key, text) {
+      ok(find(`${element}:contains(${text})`).length, `Found translation key ${key} in ${element}`);
+    };
+  } else if (typeof expect === 'function') {
+    return function(element, key, text) {
+      const found = !!(find(`${element}:contains(${text})`).length);
+      expect(found).to.equal(true);
+    };
+  } else {
+    return function() {
+      throw new Error("ember-i18n could not find a compatible test framework");
+    };
+  }
+}());
--- a/tests/.jshintrc
+++ b/tests/.jshintrc
@@ -22,7 +22,9 @@
     "andThen",
     "currentURL",
     "currentPath",
-    "currentRouteName"
+    "currentRouteName",
+    "t",
+    "expectTranslation"
   ],
   "node": false,
   "browser": false,
--- a/tests/acceptance/test-helpers-test.js
+++ b/tests/acceptance/test-helpers-test.js
@@ -1,0 +1,24 @@
+import Ember from 'ember';
+import { module, test } from 'qunit';
+import startApp from '../helpers/start-app';
+
+var app;
+
+module('Acceptance: {{t}} Helper', {
+  beforeEach: function() {
+    app = startApp();
+    visit('/');
+  },
+
+  afterEach: function() {
+    Ember.run(app, 'destroy');
+  }
+});
+
+test("t test helper", function(assert) {
+  assert.equal(t("pluralized.translation", { count: 1 }), "One Click", "test-helpers t returns translation");
+});
+
+test("expectTranslation test helper", function(assert) {
+  expectTranslation('.no-interpolations', 'no.interpolations');
+});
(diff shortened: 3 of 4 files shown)


test-support/helpers/ember-i18n/test-helpers.js
```diff
@@ -0,0 +1,15 @@
+import Ember from 'ember';
+
+// example usage: find(`.header:contains(${t('welcome_message')})`)
+Ember.Test.registerHelper('t', function(app, key){
```
jamesarosen: This should also take an optional context that it passes to `i18n.t`.

test-support/helpers/ember-i18n/test-helpers.js
```diff
@@ -0,0 +1,15 @@
...
+  var i18n = app.__container__.lookup('service:i18n');
+  return i18n.t(key);
+});
+
+// example usage: expectTranslation('.header', 'welcome_message');
+Ember.Test.registerHelper('expectTranslation', function(app, element, key){
+  var i18n = app.__container__.lookup('service:i18n'),
+      translation = i18n.t(key);
+
+  return ok(find(`${element}:contains(${translation})`),  `Found translation key ${key} in ${element}`);
```
jamesarosen: Where does `ok` come from? That looks like a QUnit global. What about for people using mocha? Sadly, [ember-test-helpers](https://github.com/switchfly/ember-test-helpers) doesn't seem to provide any cross-framework assertions.

jamesarosen:
I'd love to see a test for these as well:
```js
// tests/acceptance/test-helpers-test.js
test("t test helper", function(assert) {
  assert.equal(t("pluralized.translation", { count: 1 }), "One Click");
});

test("expectTranslation test helper" function(assert) {
  expectTranslation('.no-interpolations', 'no.interpolations');
});
```

test-support/helpers/ember-i18n/test-helpers.js
jamesarosen: @rwjblue you work on [ember-mocha](https://github.com/switchfly/ember-mocha). Do you have any suggestions for how to write this in a test-framework-agnostic way?

iezer:
@jamesarosen Thanks for your feedback. I've added a couple tests, and the `interpolations` param to the helpers. I'm not an expert on Mocha but I believe we could support both QUnit and Mocha doing something like this:

```js
if (QUnit) {
  return ok(find(`${element}:contains(${translation})`),  `Found translation key ${key} in ${element}`);
} else {
  return expect(find(`${element}:contains(${translation})`).to.equal(true);
}
```

I don't have a good answer for supporting QUnit 2.0 when they remove to global assertions.

jamesarosen:
That might work. Though there are a couple of things we might want to do to make that safer:

 1. change `if (QUnit)` to `if (typeof QUnit !== "undefined")` or even `if (typeof QUnit !== "undefined" && typeof ok !== "undefined"`
 2. have an `else if` for the mocha case. Maybe guard on `typeof expect === 'function'`
 3. the `else` case be "throw new Error("ember-i18n could not find a compatible testing framework");`

Next, instead of doing those checks each time the test helper is invoked, maybe we should pick an implementation once:

```js
const assertTranslation = (function() {
  if (typeof QUnit !== 'undefined' && typeof ok === 'function') {
    return function(element, text) {
      ok(find(`${element}:contains(${translation})`);
    };
  } else if (typeof expect === 'function') {
    return function(element, text) {
      return expect(find(`${element}:contains(${translation})`).to.equal(true);
    };
  } else {
    return function() {
      throw new Error("ember-i18n could not find a compatible test framework");
    };
  }
}());
```

And then call `assertTranslation` in `expectTranslation`.

Last, I'm a bit worried about `ok(find(...))`. `find` will return a selector, which will always be truthy. Maybe you meant `ok(find(...).length)`. Or maybe you meant `ok(find(selector).is(`:contains(${translation})`))`.

iezer:
@jamesarosen Thanks, that's a great solution. I didn't mean to return the selector. I pushed another commit with that change. Let me know if this all looks ok and I'll squash the commits into 1 for merging.

test-support/helpers/ember-i18n/test-helpers.js
```diff
@@ -0,0 +1,32 @@
...
+// example usage: find(`.header:contains(${t('welcome_message')})`)
+Ember.Test.registerHelper('t', function(app, key, interpolations){
+  var i18n = app.__container__.lookup('service:i18n');
+  return i18n.t(key, interpolations);
+});
+
+// example usage: expectTranslation('.header', 'welcome_message');
+Ember.Test.registerHelper('expectTranslation', function(app, element, key, interpolations){
+  var i18n = app.__container__.lookup('service:i18n'),
+      text = i18n.t(key, interpolations);
```
jamesarosen: This *could* reuse the `t` helper above:
```js
var text = app.testHelpers.t(key, interpolations);
```
I don't feel strongly about this.

jamesarosen:
`var` should be `const` throughout. Other than that +1 and thanks for contributing! :smile_cat: :balloon:

iezer:
@jamesarosen Thanks, changed `var` to `const` and squashed.

jamesarosen:
I dig :)

PR #276, opened 2015-07-20
Match ember-cli conventions for file structure.

blimmer:
This addon morphed to an ember-cli addon, and didn't really follow the best practices of what the file structure of most ember addons look like.

This changeset does not affect any of the external facing APIs available to users.

PR #277, opened 2015-07-20
Implement i18n.exists.

### LINKED ISSUE #264: RFC: add i18n.exists(key)
## Problem
There's no way to check for a key that may or may not be present. For example, I automatically try to assign the page title from a `pageTitles` object in my translations file. E.g.
```js
// translations.js
{
  'pageTitles': {
    'index': 'My Homepage',
    ...
  }
}
```
Some pages won't have a title token, and that's OK. In v3, I used the [exists](https://github.com/jamesarosen/ember-i18n/blob/v3.1.1/lib/i18n.js#L110) method to make sure I didn't end up with `Missing translation` messages in my page title.

## Proposal
We restore the behavior on the i18n service to check if `i18n.t(key)` would return a value. It should have no special logic outside of returning `true` if `i18n.t` returns a key from the translations file and `false` if it would fire the `missing` event.

blimmer:
Resolves #264


jamesarosen:
Blocked on #278

blimmer:
Since it looks like the #276 is going to be good to go, I rebased this based on the assumption that #276 will be merged, followed by this.

blimmer:
I also rebased this one to make sure everything's passing.

jamesarosen:
Cherry-picked as cedc6ce9099cee06e10c377e569960d2ac605917

Issue #281, opened 2015-07-23
Locales created using addTranslations don't appear in locales array


grantovich:
The `locales` array seems to only include locales that are defined in `app/locales` modules. If you create a locale at runtime by calling `addTranslations` (e.g. if you are [loading translations from an API](https://github.com/jamesarosen/ember-i18n/wiki/Example:-Fetching-Translations-Live)), it will not show up in the array (e.g. if you are [displaying a language picker](https://github.com/jamesarosen/ember-i18n/wiki/Example:-Language-Chooser)). As a workaround I'll probably hardcode the list of locales in the language picker, but it would be nice to not have this knowledge duplicated between server and client.

jamesarosen:
Nice catch. We'll probably have to do something like
```js
locales: computed(getLocales),

addTranslations(locale, translations) {
  addLocale(this.get('locales'), locale);
  ...
}

function addLocale(list, entry) {
  if (list.indexOf(entry) > -1) { return; }
  list.pushObject(entry);
}
```

kellyselden:
@grantovich I submitted a fix for this in #282.

jamesarosen:
Resolved in master

grantovich:
Sweet, thanks @kellyselden for the fix!

PR #282, opened 2015-07-28
adds a runtime locale to the array of locales on disk

kellyselden:
addresses #281

### DIFF
--- a/tests/unit/utils/locale-test.js
+++ b/tests/unit/utils/locale-test.js
@@ -4,16 +4,51 @@ moduleFor('service:i18n', 'I18nService#locales', {
   integration: true
 });
 
-const locales = [
-  'en-bw',
-  'en-ps',
-  'en-wz',
-  'en',
-  'es'
-];
-
 test('locales have the correct length and names', function(assert) {
+  const locales = [
+    'en-bw',
+    'en-ps',
+    'en-wz',
+    'en',
+    'es'
+  ];
+
   const i18n = this.subject();
+
+  assert.deepEqual(i18n.get('locales'), locales);
+});
+
+test('addTranslations adds to locales', function(assert) {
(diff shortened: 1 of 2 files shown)


jamesarosen:
Looks like we're still blocked on ember-beta and -canary :/

jamesarosen:
Blocked by #284

kellyselden:
rebased and passing

tests/unit/utils/locale-test.js
```diff
@@ -4,16 +4,51 @@ moduleFor('service:i18n', 'I18nService#locales', {
...
+    'es',
+    'it'
+  ];
+
+  const i18n = this.subject({ locale: 'en' });
+
+  i18n.addTranslations('it', {});
+  i18n.addTranslations('it', {});
+
+  assert.deepEqual(i18n.get('locales'), locales);
```
jamesarosen: I don't understand how this test is passing.
```js
x = Ember.A()
x.pushObject('it')
x.pushObject('it')
x // [ 'it', 'it' ]
```

jamesarosen:
We should probably change `get-locales` to start with an [`Ember.Set`](http://emberjs.com/api/classes/Ember.Set.html) and use `add()` instead of `pushObject()`.

tests/unit/utils/locale-test.js
kellyselden: I'm using addObject, it does the duplicate ignoring for me.

kellyselden:
@jamesarosen I'm not using `pushObject`. Also what do you mean by "start with an `Ember.set`"?

jamesarosen:
I didn't catch the distinction between `addObject` and `pushObject`. Nice.

kellyselden:
Thanks.

PR #285, opened 2015-07-28
fixing tests on beta and canary

kellyselden:


jamesarosen:
Also +1, but #276 beat you to it!

PR #306, opened 2015-09-04
Add IE8 support by changing ES6 class to a simple "class". Fixes #305

### LINKED ISSUE #305: "Object doesn't support this action" in IE8
It appears that this add-on supports Ember 1.3 but it fails in Internet Explorer 8 with the error `Object doesn't support this action`. Adding or removing the `{{t 'key'}}` helper or translation files has no effect; just adding `ember-i18n` to the package.json will take the app down.

Does anybody have an idea of where there might be incompatible code? I've been digging for a while and haven't found anything but `Object.keys` which is polyfilled.

jerel:
When an ES6 class is used Babel transpiles it with defineProperty which IE8 is can only use on DOM. By switching back to a simple class i18n is then compatible with IE8 as long as `Object.keys` is polyfilled.

### DIFF
--- a/addon/utils/locale.js
+++ b/addon/utils/locale.js
@@ -3,7 +3,7 @@ import Ember from "ember";
 // @private
 //
 // This class is the work-horse of localization look-up.
-export default class Locale {
+export default function Locale(id, container) {
 
   // On Construction:
   //  1. look for translations in the locale (e.g. pt-BR) and all parent
@@ -12,18 +12,15 @@ export default class Locale {
   //     and use the first value for `rtl` and `pluralForm`
   //  3. Default `rtl` to `false`
   //  4. Ensure `pluralForm` is defined
-  constructor(id, container) {
-    this.id = id;
-    this.container = container;
-    this.rebuild();
-  }
+  this.id = id;
+  this.container = container;
 
-  rebuild() {
+  this.rebuild = function() {
     this.translations = getFlattenedTranslations(this.id, this.container);
     this._setConfig();
-  }
+  };
 
-  _setConfig() {
+  this._setConfig = function() {
     walkConfigs(this.id, this.container, (config) => {
       if (this.rtl === undefined) { this.rtl = config.rtl; }
       if (this.pluralForm === undefined) { this.pluralForm = config.pluralForm; }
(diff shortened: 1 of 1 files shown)


jamesarosen:
v4 of this library requires Ember 1.12+. Is that version of Ember meant to be compatible with IE8?

addon/utils/locale.js
```diff
@@ -12,18 +12,15 @@ export default class Locale {
...
   //  4. Ensure `pluralForm` is defined
-  constructor(id, container) {
-    this.id = id;
-    this.container = container;
-    this.rebuild();
-  }
+  this.id = id;
+  this.container = container;
 
-  rebuild() {
```
jamesarosen: These should be on the prototype so all instances can share one copy of the functions.

jerel:
@jamesarosen this is ready for review. Git sort of mangles the diff on the last commit due to indentation but literally the only changes are replacing `this.fn` with `Locale.prototype.fn`

mattbearman:
I've upgraded to v4.2.2, but I'm still seeing `Object.defineProperty` in i18n related code after I re-build my ember app:

`/helpers/t` `services/i18n` `utils/i18n/compile-template`, `utils/i18n/missing-message`

Is there anything else I need to do to make i18n IE8 compatible?

Thanks

mattbearman:
I've found what's causing this, I'm not sure if the issue is with my set up (Ember 1.13.15), but it seems the problem occurs when i18n does `export { default } from $SOMETHING`. Eg: in `node_modules/ember-i18n/app/helperst.js` you have:

```javascript
export { default } from 'ember-i18n/helper';
```

which gets compiled into

```javascript
define('my-ember-app/helpers/t', ['exports', 'ember-i18n/helper'], function (exports, _emberI18nHelper) {
  Object.defineProperty(exports, 'default', {
    enumerable: true,
    get: function get() {
      return _emberI18nHelper['default'];
    }
  });
});
```

I've found that by creating my own helper file in `my-ember-app/app/helpers/t.js` with the more verbose:

```js
import t from 'ember-i18n/helper';

export default t;
```

The compiled version becomes a much more IE8 friendly:

```js
define('my-ember-app/helpers/t', ['exports', 'ember-i18n/helper'], function (exports, _emberI18nHelper) {
  exports['default'] = _emberI18nHelper['default'];
});
```

The helper still seems to work fine, I'm now going to try this for the service and utilities to see if it gets my app working in IE8.

Damn I wish I didn't need to support IE8

jamesarosen:
Do you want to open a PR for the import-then-export change?

Issue #298, opened 2015-08-31
Translate into a locale other than the current one


grantovich:
`i18n` gem has this:

```
I18n.t('some.key', locale: 'en')
```

It would be nice to have something similar for `ember-i18n`. For instance, I have a locale picker that displays each language as that language's name for itself (i.e. English, Español, Français). To have this display correctly in all languages, I need to duplicate that list of languages in all of my locales. It would be cleaner if each locale only stored its own name, and I could call `t('name', { locale: 'en' })`.

(technically that approach would be backward-incompatible with any translations that legitimately had an interpolation named `locale` in them, but that is how the `i18n` gem handles it)

jamesarosen:
Can you link to the relevant Ruby?

grantovich:
This is the part of the `i18n` gem that picks up the locale from the `translate` options:

https://github.com/svenfuchs/i18n/blob/b1200f0289a59ba0368f800473d8ac79f713bc26/lib/i18n.rb#L148

Strangely the method's documentation does not mention this behavior, but the docs for the `transliterate` method immediately below it do include an example.

jamesarosen:
This seems like a fairly unlikely key for a translation to have. Still, it might be good to introduce a flag. Something like

```js
i18nService.set('CROSS_LOCALE_LOOKUPS', true);
```

I would suggest three options:

 1. `warn`: treat `locale` as a special key, warn if a translation uses it; default
 1. `true`: treat `locale` as a special key, silence warnings
 1. `false`: treat `locale` as a regular key, silence warnings

grantovich:
Sounds good to me, although I imagine you'd eventually want to remove the warning/option and have `locale` always be special (perhaps in the next major version bump). Since there is `ENV.i18n` already, should the option go there?

jamesarosen:
> I imagine you'd eventually want to remove the warning/option and have locale always be special (perhaps in the next major version bump)

Exactly.

> Since there is ENV.i18n already, should the option go there?

Sure. The `service:i18n` can pull it off there.

jamesarosen:
jamesarosen/ember-i18n has been deprecated in favor of [ember-intl](https://github.com/ember-intl).

Issue #255, opened 2015-06-24
RFC: support for translations in addons


jamesarosen:
In the discussion below, I will use two hypothetical addons, _ember-paginate_ and _ember-select_.
## Problem

Currently, if an addon wants to provide default translations that the app can consume and optionally override, there are two options:
### Require App to Manage

Each addon has an `addon/locales/en/translations.js` file and similar files for all locales it supports. The app must import these in its own translations files:

``` js
// my-app/app/locales/en/translations.js:

import paginationTranslations from "ember-paginate/locales/en/translations";
import selectTranslations from "ember-select/locales/en/translations";
import Ember from "ember";

const translations = {};
Ember.merge(translations, paginationTranslations);
Ember.merge(translations, selectTranslations);

Ember.merge(translations, {
  "app.title": "My App"
});

export default translations;
```

That's a lot of boilerplate, especially since it needs to be repeated for every locale.
### Addon Injects Translations

The addon could inject translations at runtime with an instance-initializer:

``` js
// ember-paginate/app/instance-initializers/ember-paginate.js:
import en from "ember-paginate/locales/en/translations.js";
import fr from "ember-paginate/locales/fr/translations.js";
import zh from "ember-paginate/locales/zh/translations.js";

export default {
  name: 'ember-paginate',
  initialize: function({ container }) {
    const i18n = container.lookup('service:i18n');
    i18n.addTranslations("en", en);
    i18n.addTranslations("fr", fr);
    i18n.addTranslations("zh", zh);
  }
};
```

That's far better from an app-developer standpoint, but worse for addon authors.
## Goal

The goal of this RFC is to make it so easy for both addon- and app developers to support i18n that they do by default.
## Proposal

The key to this implementation is to loosen the restriction on the names of translation files. _ember-paginate_ would have `ember-paginate/app/locales/en/ember-paginate.js`. _ember-select_ might have `ember-select/app/locales/en/ember-select.js`.

It would then be the responsibility of ember-i18n's `Locale` to merge translations from all of the various packages. The new logic to get the collected translations for locale `ab-CDE` would be
1. enumerate `locale:ab-CDE/*`, excluding `locale:ab-CDE/config`
2. merge those packages in alphabetical order
3. walk up the locale-specificity to `ab` and repeat

https://github.com/emberjs/ember.js/pull/11393 introduced a new private API, `knownForType`, that would be useful for enumerating these packages.
### Ancillary Benefits

[...]

PR #312, opened 2015-09-11
Allow overriding the locale for individual translations

grantovich:
I had some free time, so here is my initial stab at #298. I couldn't find a way to stub the ENV value so I could test that it's correctly loaded into the service, but I manually confirmed that it works.

### DIFF
--- a/addon/services/i18n.js
+++ b/addon/services/i18n.js
@@ -18,21 +18,20 @@ export default Parent.extend(Evented, {
   locales: computed(getLocales),
 
   // @public
+  // Whether to allow overriding the locale for individual translations.
+  allowLocaleOverride: undefined,
+
+  // @public
   //
   // Returns the translation `key` interpolated with `data`
   // in the current `locale`.
   t(key, data = {}) {
-    Ember.deprecate('locale is a reserved attribute', data['locale'] === undefined, {
-      id: 'ember-i18n.reserve-locale',
-      until: '5.0.0'
-    });
-
     Ember.deprecate('htmlSafe is a reserved attribute', data['htmlSafe'] === undefined, {
       id: 'ember-i18n.reserve-htmlSafe',
       until: '5.0.0'
     });
 
-    const locale = this.get('_locale');
+    const locale = this._getLocaleWithOverride(data);
     assert("I18n: Cannot translate when locale is null", locale);
     const count = get(data, 'count');
 
@@ -42,7 +41,7 @@ export default Parent.extend(Evented, {
     const template = locale.getCompiledTemplate(defaults, count);
 
     if (template._isMissing) {
-      this.trigger('missing', this.get('locale'), key, data);
+      this.trigger('missing', locale.id, key, data);
     }
 
     return template(data);
@@ -50,7 +49,7 @@ export default Parent.extend(Evented, {
 
   // @public
   exists(key, data = {}) {
-    const locale = this.get('_locale');
+    const locale = this._getLocaleWithOverride(data);
     assert("I18n: Cannot check existance when locale is null", locale);
     const count = get(data, 'count');
 
@@ -83,6 +82,10 @@ export default Parent.extend(Evented, {
       }
       this.set('locale', defaultLocale);
     }
+
+    if (this.get('allowLocaleOverride') == null) {
+      this.set('allowLocaleOverride', (ENV.i18n || {}).allowLocaleOverride);
+    }
   }),
 
   // @private
@@ -96,5 +99,24 @@ export default Parent.extend(Evented, {
     const locale = this.get('locale');
 
     return locale ? new Locale(this.get('locale'), getOwner(this)) : null;
-  })
+  }),
+
+  // @private
+  // Retrieves the current locale, allowing it to be overridden by a `locale`
+  // option in the translation data if `allowLocaleOverride` is enabled.
+  _getLocaleWithOverride: function(data) {
+    let locale = this.get('_locale');
+
+    if (data.locale && this.get('allowLocaleOverride') !== false) {
+      Ember.warn(
+        'Specifying an interpolation key named `locale` now overrides the target locale for translation. To silence this warning, set ENV.i18n.allowLocaleOverride to true. To disable this behavior and treat `locale` as a normal interpolation key, set ENV.i18n.allowLocaleOverride to false.',
+        (this.get('allowLocaleOverride') !== undefined),
+        { id: 'ember-i18n.locale-override' }
+      );
+      locale = new Locale(data.locale, getOwner(this));
+      delete data.locale;
+    }
+
+    return locale;
+  }
 });
--- a/tests/unit/i18n-t-test.js
+++ b/tests/unit/i18n-t-test.js
@@ -1,13 +1,23 @@
 import Ember from 'ember';
 import { moduleFor, test } from 'ember-qunit';
 
-const { run, deprecate:originalDeprecate } = Ember;
+const { run, deprecate:originalDeprecate, warn:originalWarn } = Ember;
+const warnings = Ember.A();
 
 moduleFor('service:i18n', 'I18nService#t', {
   integration: true,
 
+  beforeEach() {
+    Ember.warn = function(message, test, options) {
+      if (!test) { warnings.pushObject(message); }
+      originalWarn(message, test, options);
+    };
+  },
+
   afterEach() {
     Ember.deprecate = originalDeprecate;
+    Ember.warn = originalWarn;
+    warnings.clear();
   }
 });
 
@@ -40,7 +50,7 @@ test('returns "missing translation" translations', function(assert) {
   assert.equal('Missing translation: not.yet.translated', result);
 });
 
-test('warns on the presence of htmlSafe and locale', function(assert) {
+test('warns on the presence of htmlSafe', function(assert) {
   const service = this.subject();
   let deprecations = 0;
 
@@ -49,7 +59,7 @@ test('warns on the presence of htmlSafe and locale', function(assert) {
 
     const { id } = options;
 
-    if (id === 'ember-i18n.reserve-htmlSafe' || id === 'ember-i18n.reserve-locale') {
+    if (id === 'ember-i18n.reserve-htmlSafe') {
       deprecations += 1;
     }
   };
@@ -57,12 +67,9 @@ test('warns on the presence of htmlSafe and locale', function(assert) {
   service.t('not.yet.translated', { htmlSafe: true });
   assert.equal(deprecations, 1);
 
-  service.t('not.yet.translated', { locale: true });
-  assert.equal(deprecations, 2);
-
   service.t('not.yet.translated');
   service.t('not.yet.translated', { some: 'other key' });
-  assert.equal(deprecations, 2);
+  assert.equal(deprecations, 1);
 });
 
 test('emits "missing" events', function(assert) {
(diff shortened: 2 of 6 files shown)


addon/services/i18n.js
```diff
@@ -19,11 +19,23 @@ export default Parent.extend(Ember.Evented, {
...
   // Returns the translation `key` interpolated with `data`
   // in the current `locale`.
   t: function(key, data = {}) {
-    const locale = this.get('_locale');
+    let locale = this.get('_locale');
+    if (data.locale && this.get('allowLocaleOverride') !== false) {
+      if (this.get('allowLocaleOverride') === undefined) {
+        Ember.warn('Specifying an interpolation key named `locale` now overrides the target locale for translation. To silence this warning, set ENV.i18n.allowLocaleOverride to true. To disable this behavior and treat `locale` as a normal interpolation key, set ENV.i18n.allowLocaleOverride to false.');
+      }
+      locale = new Locale(data.locale, this.container);
```
grantovich: This part seems inefficient, but addressing it would probably require a larger change (for instance, locales could be cached on the service when they are initialized, or all locales could be loaded up-front).

addon/services/i18n.js
```diff
@@ -19,11 +19,26 @@ export default Parent.extend(Ember.Evented, {
...
-    const locale = this.get('_locale');
+    let locale = this.get('_locale');
+    if (data.locale && this.get('allowLocaleOverride') !== false) {
+      if (this.get('allowLocaleOverride') === undefined) {
+        Ember.warn(
+          'Specifying an interpolation key named `locale` now overrides the target locale for translation. To silence this warning, set ENV.i18n.allowLocaleOverride to true. To disable this behavior and treat `locale` as a normal interpolation key, set ENV.i18n.allowLocaleOverride to false.',
+          { id: 'ember-i18n.locale-override' }
+        );
+      }
+      locale = new Locale(data.locale, this.container);
```
grantovich: This part seems inefficient, but addressing it would probably require a larger change (for instance, locales could be cached on the service when they are initialized).

addon/services/i18n.js
```diff
@@ -19,11 +19,25 @@ export default Parent.extend(Ember.Evented, {
...
   t: function(key, data = {}) {
-    const locale = this.get('_locale');
+    let locale = this.get('_locale');
+    if (data.locale && this.get('allowLocaleOverride') !== false) {
+      Ember.warn(
+        'Specifying an interpolation key named `locale` now overrides the target locale for translation. To silence this warning, set ENV.i18n.allowLocaleOverride to true. To disable this behavior and treat `locale` as a normal interpolation key, set ENV.i18n.allowLocaleOverride to false.',
+        (this.get('allowLocaleOverride') !== undefined),
+        { id: 'ember-i18n.locale-override' }
+      );
+      locale = new Locale(data.locale, this.container);
```
grantovich: This part seems inefficient, but addressing it would probably require a larger change (for instance, locales could be cached on the service when they are initialized).

app/instance-initializers/ember-i18n.js
```diff
@@ -14,6 +14,7 @@ export default {
     const key = 'service:i18n';
     const i18n = instance.lookup ? instance.lookup(key) : instance.container.lookup(key);
     i18n.set('locale', defaultLocale);
+    i18n.set('allowLocaleOverride', (ENV.i18n || {}).allowLocaleOverride);
```
jamesarosen: Because this initializer doesn't run in all situations (notably non-integration tests), I suggest moving this logic into the Service itself:
```js
allowLocaleOverride: Ember.computed({
  get() { return (ENV.i18n || {}).allowLocaleOverride; },
  set(prop, newValue) { return newValue; }
})
```

tests/unit/i18n-t-test.js
```diff
@@ -71,3 +88,27 @@ test("check unknown locale", function(assert) {
...
+  assert.equal(warnings.length, 0);
+});
+
+test("locale can be used as an interpolation key if allowLocaleOverride is false", function(assert) {
+  const i18n = this.subject({ locale: 'en', allowLocaleOverride: false });
+
+  const result = i18n.t('with.locale.interpolation', { locale: 'es' });
+  assert.equal(result.string, 'Locale: es');
+  assert.equal(warnings.length, 0);
+});
```
jamesarosen: These tests are awesome :)

addon/services/i18n.js
jamesarosen: 1. we only want to warn if `data.locale` is defined
 2. we can make it slightly more efficient by saving off the warning text as a `const` at the top of the file
 3. we could cache `Locale`s in a closure-local map in the `locale.js` file:

```js
const seenLocales = {};

function Locale(name, container) {
  if (seenLocales[name]) { return seenLocales[name]; }
  seenLocales[name] = this;
  ...
}
```

grantovich: > we only want to warn if `data.locale` is defined

This is inside the `if (data.locale && ...)` so the warning should only be seen in that case.

> we could cache `Locale`s in a closure-local map in the `locale.js` file

Clever! I will see if this works.

jamesarosen: > This is inside the if (data.locale && ...)

d'oh

grantovich: Though I can't help thinking that from the outside, it's very counter-intuitive that `new Locale` only *really* constructs a new locale the first time it's called with a given argument, and thereafter just returns the previously-constructed object. At that point it's really more like a lazily-populated hash. Maybe a "locales" service or something?

jamesarosen: That file is private, so we can change its API. We can change the exported function to `localeFor(name, container)` and uses the map.

app/instance-initializers/ember-i18n.js
grantovich: I think it's probably a good idea that all the code which initializes the service with ENV values should be in the same place – it would be strange to have one property over here, and one property over there. Or would you also suggest moving the `defaultLocale` stuff? (that seems like a separate change)

addon/services/i18n.js
grantovich: Sounds like a good simple solution, I will try going in that direction.

tests/unit/i18n-t-test.js
```diff
@@ -71,3 +88,27 @@ test("check unknown locale", function(assert) {
...
+  const result = i18n.t('no.interpolations', { locale: 'es' });
+  assert.equal(result.string, 'texto sin interpolaciones');
+  assert.equal(warnings.length, 0);
+});
+
+test("locale can be used as an interpolation key if allowLocaleOverride is false", function(assert) {
+  const i18n = this.subject({ locale: 'en', allowLocaleOverride: false });
+
+  const result = i18n.t('with.locale.interpolation', { locale: 'es' });
+  assert.equal(result.string, 'Locale: es');
```
grantovich: For some reason, in these tests alone, the `t` function seems to return an object with a single property called `'string'`, which I have to access to get the actual translation. Really not sure why that is.

app/instance-initializers/ember-i18n.js
jamesarosen: Yes, I think that makes sense as well. I'll do that now and link the PR.

jamesarosen: Except the service is in `addon/` and the config is in `app/`. Hmm. I don't know that there's a way to reference `ENV` from the service directly :(

grantovich: I think it's fine – this way the service itself is isolated, and doesn't try to pull in outside information when you unit-test it. Presumably the initializer logic could be covered by an integration test (if there was a way to stub ENV values).

jamesarosen: I meant that in an app's unit tests, they're not going to get the same behavior that they get in an integration test. It's going to fail in a way that's hard to diagnose.

grantovich: Ah, I see, you were talking about the *user's* integration tests and unit tests. My bad.

jamesarosen: From elwayman02/ember-data-echonest:

```
const ENV = this.container.lookupFactory('config:environment');
```

jamesarosen: You may want to rebase off of master once #313 is merged.

grantovich:
@jamesarosen I tried changing things to use your `localeFor`-style proposal, but that caused test failures related to adding translations at runtime. Strangely, using your originally proposed approach (returning from the constructor) does work.

I've addressed the outstanding feedback and rebased. The `.string` issue seems to have mysteriously disappeared... so I guess this should be good to go.

grantovich:
@jamesarosen Any further changes needed? Could this be merged and released? (I guess it should have a changelog entry, I could write that if you like)

grantovich:
When testing this in my app, I discovered a need to find whether a given translation exists in a locale other than the current one. So I extracted `_getLocaleWithOverride` and applied it to `exists` as well. Not sure what's up with the test failures on Travis – I don't get them locally, and they seem unrelated to my changes.

addon/utils/locale.js
```diff
@@ -13,6 +15,10 @@ export default class Locale {
   //  3. Default `rtl` to `false`
   //  4. Ensure `pluralForm` is defined
   constructor(id, container) {
+    if (initializedLocales[id]) {
+      return initializedLocales[id];
+    }
```
jamesarosen: You never populate `initializedLocales` with `this` after the `if`.

addon/services/i18n.js
```diff
@@ -41,7 +46,7 @@ export default Parent.extend(Evented, {
 
   // @public
   exists: function(key, data = {}) {
-    const locale = this.get('_locale');
+    const locale = this._getLocaleWithOverride(data);
```
jamesarosen: I like the symmetry of `t` and `exists` now. I'm glad you did that extraction.

addon/utils/locale.js
grantovich: Argh. You're completely right. Unfortunately, actually doing this causes adding translations at runtime to stop working. So I guess this method of caching won't work after all. I'm really not sure how to handle this correctly, the architecture is hard for me to follow... should I just take it out for now? I can personally live with the sub-optimal performance, since my use case is not very intensive (this may be a case of premature optimization).

jamesarosen: Can you try removing the `if` around `locale.rebuild()` in `addTranslations`?

I think we need to do some substantial re-architecting anyway to support #255. Probably best to think about the two problems together.

grantovich: Just tried that, but the test for adding translations at runtime still fails.

jamesarosen: OK. I'll give #255 a shot first. I may try moving some of the `addTranslations` logic into `Locale` so the `I18nService` doesn't have to worry about state management.

grantovich: Alright. For now I'll proceed with using this branch in my app, since it seems stable and I haven't noticed any performance hits (I have a small number of translations, and the cross-locale lookup is only performed once on page load). I will keep an eye on that task.

grantovich:
@jamesarosen I've resolved conflicts with current master and removed the defunct attempt at caching. How would you like to handle this? Should it still wait on #255? Can it be merged with a potential caveat about performance? (I've been using this in production for a month now and haven't seen any breaking issues with it, but also haven't examined performance too closely)

jamesarosen:
Sorry I've let this sit so long. I want to work on #255 very soon. I'll try to eke out some time this week. Once I have a better picture of that work, I'll let you know what we should do with this. I definitely want either this code or this idea in the project.

grantovich:
Rebased this on v4.2.0 to resolve conflicts. @jamesarosen how's it going? :smiley:

jamesarosen:
I'm hoping to do some of this heavier work over the break right before the new year. I'll keep you updated.

grantovich:
Hi @jamesarosen, just want to confirm whether this is still slated to be merged at some point. In our limited use case we've had no problems with it, and have been using it in production since October – but we don't want to maintain a fork of ember-i18n indefinitely.

Of course, I'd be happy to write a changelog entry and wiki documentation for this feature, or any other friction-reducing tasks that would help.

jamesarosen:
There's not much movement on ember-i18n these days because I'm working on a release for another project right now. But after that release will be some I18n work, which will be a perfect time to pick this up again.

Thanks for testing it out!

sbsurf:
This would definitely be useful. I could use this as a way to inject missing translation fallbacks, while overriding the missingMessage function to just return the english translation.

sandstrom:
@jamesarosen This would be a wonderful addition! ⛵ 

Is there anything in particular holding this PR back?

sandstrom:
@grantovich Great initiative! 

What do you think about this line:
https://github.com/jamesarosen/ember-i18n/blob/9e840285b53879e5a1eeb7fca57076b69b6f8c9b/addon/services/i18n.js#L41

```js
this.trigger('missing', this.get('locale'), key, data); # current
this.trigger('missing', locale.id, key, data); # my suggestion
```

grantovich:
@sandstrom Thanks for the catch – I haven't been keeping this up-to-date lately since it seems there isn't much interest in merging it, and our project that uses this fork is in maintenance mode. If @jamesarosen gives the go-ahead, I would be happy to revisit it.

(tagging @jasonmit since from https://github.com/jamesarosen/ember-i18n/pull/381#issuecomment-230182277 it looks like we might be duplicating some effort)

jpsirois:
I just had this exact same need for a project. And I extended the i18n services with the changes from this PR and it just work. Is this something still in consideration? I think this PR is worth consideration.

sandstrom:
friendly ping @jamesarosen 😄

jamesarosen:
I don't have any objections. I'd love one of the other maintainers to take a look and help shepherd this.

jamesarosen:

PR #321, opened 2015-10-07
remove Stream awareness from service:i18n

jamesarosen:
The only time that the `service:i18n` needs to expose a `localeStream` is when we're using the legacy helper (Ember 1.12). Therefore, instead of always having the service build that Stream, we use the initializer to set it up only when we're installing the legacy helper. This makes the library completely streamless on Ember 1.13+.

### DIFF
--- a/addon/stream.js
+++ b/addon/stream.js
@@ -2,13 +2,10 @@ import Ember from 'ember';
 
 // As of v1.12, Streams are still private API. Thus, we need to reach in to
 // Ember internals to get access to it.
-// After v2.1, the streams/stream module moved what was `Stream` to a named export and exported
-// a base class as `default`.
 //
 // See https://github.com/emberjs/ember.js/blob/v1.12.0/packages/ember-metal/lib/main.js#L384-L386
 // See https://github.com/emberjs/ember.js/pull/9693
 // See https://github.com/dockyard/ember-cli-i18n/blob/v0.0.6/addon/utils/stream.js
-// See https://github.com/emberjs/ember.js/blob/23258c1eadce4f52c814f0441c13880ddf896f31/packages/ember-metal/lib/streams/stream.js
-const streamModule = Ember.__loader.require('ember-metal/streams/stream');
-export default (streamModule.Stream || streamModule['default']);
-export var readHash = Ember.__loader.require('ember-metal/streams/utils').readHash;
+
+export default Ember.__loader.require('ember-metal/streams/stream')['default'];
+export const readHash = Ember.__loader.require('ember-metal/streams/utils').readHash;
--- a/app/instance-initializers/ember-i18n.js
+++ b/app/instance-initializers/ember-i18n.js
@@ -1,12 +1,24 @@
 import Ember from "ember";
+import Stream from 'ember-i18n/stream';
 import legacyHelper from "ember-i18n/legacy-helper";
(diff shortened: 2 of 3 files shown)


app/instance-initializers/ember-i18n.js
```diff
@@ -1,12 +1,24 @@
 import Ember from "ember";
+import Stream from 'ember-i18n/stream';
 import legacyHelper from "ember-i18n/legacy-helper";
 import ENV from '../config/environment';
 
 export default {
   name: 'ember-i18n',
 
-  initialize: function(instance) {
+  initialize: function({ container }) {
```
Kuzirashi: Curiosity: what will it do? container will be applicationInstance.container? Some kind of destructuring? Taking container as argument is deprecated, but you use here `{}` so that's why I'm asking. :) @jamesarosen

addon/stream.js
```diff
@@ -2,13 +2,10 @@ import Ember from 'ember';
...
 //
 // See https://github.com/emberjs/ember.js/blob/v1.12.0/packages/ember-metal/lib/main.js#L384-L386
 // See https://github.com/emberjs/ember.js/pull/9693
 // See https://github.com/dockyard/ember-cli-i18n/blob/v0.0.6/addon/utils/stream.js
-// See https://github.com/emberjs/ember.js/blob/23258c1eadce4f52c814f0441c13880ddf896f31/packages/ember-metal/lib/streams/stream.js
-const streamModule = Ember.__loader.require('ember-metal/streams/stream');
-export default (streamModule.Stream || streamModule['default']);
+
+export default Ember.__loader.require('ember-metal/streams/stream')['default'];
 export var readHash = Ember.__loader.require('ember-metal/streams/utils').readHash;
```
Kuzirashi: If you go so much ES6, you could use `const` here. :)

jamesarosen: Yeah, but I wasn't changing this line otherwise.

app/instance-initializers/ember-i18n.js
jamesarosen: This is an `instance-initializer` not an `initializer`, so it always receives an object that looks like `{ container: AContainer, ... }`. This destructuring pulls that property off of it.

Kuzirashi: Thank you for explanation and pointing a difference.

Kuzirashi: You can also consider using ES6 shorthand for declaring function:

`initialize: function({ container }) {` 

->

`initialize({ container }) {`

jamesarosen: Good catch. We use that extensively elsewhere in this project.

app/instance-initializers/ember-i18n.js
```diff
@@ -1,12 +1,24 @@
 import Ember from "ember";
+import Stream from 'ember-i18n/stream';
 import legacyHelper from "ember-i18n/legacy-helper";
 import ENV from '../config/environment';
 
 export default {
   name: 'ember-i18n',
 
-  initialize: function(instance) {
+  initialize({ container }) {
```
rwjblue: FYI - accessing `instance.container` will trigger a deprecation in 2.1.  Moving this destructuring down into the `if` below will ensure it is only ran on pre-1.13 (where `instance.container` is fine).

Issue #322, opened 2015-10-08
Accessing translations from missing-message util


ofridagan:
It would be nice if we could access the I18n service from utils/i18n/missing-message.js
This way we could implement a default fallback language for example..

jamesarosen:
Interesting. The current signature is
```js
function missingMessage(locale, key, data)
```
We could add I18n to that:
```js
function missingMessage(locale, key, data, i18n)
```

thagul:
It would be great because with that change I could send a notification for the missing translation but also display the english version (english is my fallback).

jamesarosen:
I suppose it could alternatively be called *on the I18n service*, so you could do
```js
function missingMessage(locale, key, data) {
  return this.t('missing-translation', { key: key, locale: 'en' });
}
```
I'm not sure that's as obvious, though.

thagul:
@jamesarosen I didn't understand your last message. Are you saying that I can already do what I need?

Let me make an example of my needs. I've two language files `en` and `it`. `EN` is my base language and it's always complete, `it` may have problems. 

English file contains:
```
{
"label1": "Homepage",
"label2": "Contact us",
"label3": "About us"
}
```

Italian file contains:
```
{
"label1": "Homepage",
"label2": "Contact us"
}
```
`label3: about us` is missing (I've forgot it or I haven't translated yet, it's the same).

When the user visit my application, I don't wan't to show him a `Missing translation: label3` message but I would show him the english version. In the meantime I would receive an alert to get informed of the missing translation (I've already configured ember to send an email when an error occurs).

Any suggestion on how can do that? 
Many thanks

jamesarosen:
I was discussing alternative implementation ideas, not suggesting it was already possible. Would you be willing and able to make a PR to add `i18n` as a final argument to `missingMessage`? That area of the code is pretty well tested; it should be pretty easy to add another test.

ofridagan:
@jamesarosen yes, this is what I had in mind.
I created a PR with the simple fix: https://github.com/jamesarosen/ember-i18n/pull/323

ofridagan:
btw, even with this fix, what would be the best way to implement a default locale?
One option I see is instantiating a Locale object, and use that... but it seems very hacky.
I tried to do something like: ```i18n.t(key, {locale: 'en'})``` but it doesn't work...

jamesarosen:
See #256 for a discussion of falling back to a default locale.

PR #323, opened 2015-10-09
Calling missingMessage with the I18n service as the context

ofridagan:
Calling missingMessage with the I18n service as the context to make it possible to use I18n to implement a default translation language, for example. see: https://github.com/jamesarosen/ember-i18n/issues/322

### DIFF
--- a/addon/utils/locale.js
+++ b/addon/utils/locale.js
@@ -29,7 +29,7 @@ export default class Locale {
       if (this.pluralForm === undefined) { this.pluralForm = config.pluralForm; }
     });
 
-    const defaultConfig = this.container.lookup('ember-i18n@config:zh');
+    const defaultConfig = this.container.lookupFactory('ember-i18n@config:zh');
 
     if (this.rtl === undefined) {
       Ember.warn(`ember-i18n: No RTL configuration found for ${this.id}.`);
@@ -87,7 +87,7 @@ export default class Locale {
   }
 
   _defineMissingTranslationTemplate(key) {
-    const I18n = this.container.lookupFactory('service:i18n');
+    const I18n = this.container.lookup('service:i18n');
     const missingMessage = this.container.lookupFactory('util:i18n/missing-message');
     const locale = this.id;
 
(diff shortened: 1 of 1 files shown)


addon/utils/locale.js
```diff
@@ -87,10 +87,11 @@ export default class Locale {
   }
 
   _defineMissingTranslationTemplate(key) {
+    const I18n = this.container.lookupFactory('service:i18n');
```
jamesarosen: You want `lookup`, not `lookupFactory`.

addon/utils/locale.js
```diff
@@ -87,10 +87,11 @@ export default class Locale {
...
     const missingMessage = this.container.lookupFactory('util:i18n/missing-message');
     const locale = this.id;
 
-    function missingTranslation(data) { return missingMessage(locale, key, data); }
+    function missingTranslation(data) { return missingMessage.call(I18n, locale, key, data); }
```
jamesarosen: Since we're going to call `missingMessage` on the` service:i18n`, why don't we just add it to that object's public API:
```js
// services/i18n:
{
  buildMissingTranslation(locale, key, data) {
    const fn = this.container.lookupFactory('util:i18n/missing-message');
    return fn.call(this, locale, key, data);
  }
}

// locale.js:
{
  _defineMissingTranslationTemplate(key) {
    ...
    function missingTranslation(data) {
      return i18n.buildMissingTranslation(locale, key, data);
    }
    ...
  }
}
```

ofridagan: I'm probably not nearly familiar enough with the i18n API... but I don't see the point in allowing to call buildMissingTranslation on i18n object.
BTW, won't this cause a recursion loop if called from the missing-message function, i.e:
```export default function(locale, key, data) {
  return this.buildMissingTranslation(locale, key, data)
}```

jamesarosen: The idea was to consolidate the public API. Currently, the way to override the functionality is to define a `utils/i18n/missing-message.js` in your app. If it becomes a public method of `service:i18n`, then it, like every other override, happens in `app/services/i18n.js`. But I don't think it's a strong requirement. We can always move it later.

jamesarosen:
This is better than what we have. Merging. We can think about moving the API later.

PR #325, opened 2015-10-22
Update pl.js

jembezmamy:
Polish pluralization rules are not the same as russian. :D
I've corrected them according to http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html

### DIFF
--- a/addon/config/pl.js
+++ b/addon/config/pl.js
@@ -1,2 +1,16 @@
-import ru from './ru';
-export default ru;
+import { ONE, FEW, MANY, OTHER } from './constants';
+
+export default {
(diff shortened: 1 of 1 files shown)


pjmorse:
I think this looks :+1: .

As an aside, I wonder if a low-priority task might be to write a test which loads each of the locale files in turn (or at least the ones which define their own `pluralForm()`) and throws a few hundred numbers at `pluralForm()` just to validate it doesn't do anything unexpected.

jamesarosen:
Beautiful interpretation of the CLDR rules :)

PR #336, opened 2015-11-14
Use public API for container / registry functions.

rwjblue:
As of Ember 2.3, using `this.container` from within an instance that is looked up from the container will issue a deprecation warning suggesting that you should refactor away from the private `this.container` usage to the new public API for using container/registry features (`Ember.getOwner(this)`).

This change pulls in the `ember-getowner-polyfill` to allow us to use `getOwner` on all Ember versions (tested back to 1.10 in that addon's CI).

### DIFF
--- a/addon/services/i18n.js
+++ b/addon/services/i18n.js
@@ -1,4 +1,5 @@
 import Ember from "ember";
+import getOwner from 'ember-getowner-polyfill';
 import Locale from "../utils/locale";
 import addTranslations from "../utils/add-translations";
 import getLocales from "../utils/get-locales";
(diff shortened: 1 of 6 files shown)


rwjblue:
I noticed that Ember 2.0 wasn't under CI so I submitted https://github.com/jamesarosen/ember-i18n/pull/337 with some ember-try config cleanup.

rwjblue:
Ember 2.3.0-beta.1 has been released, and anyone using ember-i18n prior to this PR will be receiving deprecations.

jamesarosen:
+1

PR #337, opened 2015-11-14
Update ember-try configuration.

rwjblue:
* Add Ember 2.0.0 to CI (not currently being tested because release is 2.1).
* Add Ember 2.1.0 to CI explicitly (when Ember 2.2.0 ships this week 2.1 will no longer be tested by CI without this change).
* Update resolutions for earlier versions (can't use `release` as a resolution for 1.12 etc).


rwjblue:
Ember 2.2.0 is now release, so without this PR neither 2.0 or 2.1 are under test.

jamesarosen:
+1

Issue #341, opened 2015-12-03
installation process for v3.1.1 incorrectly inserts 4.x into bower.json


asross:
I have a commit [here](https://github.com/asross/ember-i18n/commit/1946bd231fbc2bb281bfe054a234bfea4bd3ff19) with an explanation of the problem and a fix, but I can't make a pull request because there's no 3.x branch.

jamesarosen:
[There is now](https://github.com/jamesarosen/ember-i18n/tree/v3)!

asross:
Thanks :)

PR #342, opened 2015-12-03
ensure installation of v3 doesn't insert v4 into bower.json

asross:
As per https://github.com/jamesarosen/ember-i18n/issues/341


jamesarosen:
+1

PR #363, opened 2016-04-18
Replaced usage of `new Ember.Handlebars.SafeString` with `Ember.String.htmlSafe`

workmanw:
Using `SafeString` in this way has been deprecated. See: http://emberjs.com/deprecations/v2.x/#toc_use-ember-string-htmlsafe-over-ember-handlebars-safestring

`Ember.String.htmlSafe` has been around since Ember 1.0.0 so I don't see any concerns about backwards compatibility.

### DIFF
--- a/addon/utils/i18n/compile-template.js
+++ b/addon/utils/i18n/compile-template.js
@@ -1,6 +1,6 @@
 import Ember from "ember";
 
-const SafeString = Ember.Handlebars.SafeString;
+const htmlSafe = Ember.String.htmlSafe;
 const get = Ember.get;
 const escapeExpression = Ember.Handlebars.Utils.escapeExpression;
 const tripleStache = /\{\{\{\s*(.*?)\s*\}\}\}/g;
@@ -21,6 +21,6 @@ export default function compileTemplate(template, rtl = false) {
 
     const wrapped = rtl ? `\u202B${result}\u202C` : result;
 
-    return new SafeString(wrapped);
+    return htmlSafe(wrapped);
   };
 }


workmanw:
Obviously there is a failing test here. The TL;DR; is that we can no longer test with `instanceof Ember.Handlebars.SafeString`. I've submitted an issue to Ember: https://github.com/emberjs/ember.js/issues/13354 . Pending the outcome of that, I'll circle back around fix this.

jamesarosen:
Build now failing on beta: https://travis-ci.org/jamesarosen/ember-i18n/jobs/126707715

workmanw:
@jamesarosen Sorry for the delay. I circled back around this morning on the emberjs issue. There is now a PR open to remove the deprecation: https://github.com/emberjs/ember.js/pull/13442 . When that is merged and a new beta release is cut, the tests should pass once again. It seems the plan is still to switch to `Ember.String.htmlSafe`, so once the next build of emberjs-beta drops, I would advocate we merge this.

devotox:
The new beta has dropped should this be retested?

workmanw:
I'm away from my computer until tonight. Once I get home I'll force push a
new commit to kick-off a new set of test runs. Sorry for the delay!

workmanw:
@devotox @jamesarosen Looks like the tests pass now. I think this should still be merged even though using `new Ember.Handlebars.SafeString` is no longer deprecated.

jamesarosen:
What about the new _is-safe_ implementation that we use in the tests?

workmanw:
@jamesarosen That's still in the RFC phase. Unfortunately there has been little activity there. It works now because `htmlSafe` will still return an instanceof `Ember.Handlebars.SafeString`. 

Maybe the better path would be to hold on this until that RFC goes through and this feature lands.

jamesarosen:
> Maybe the better path would be to hold on this until that RFC goes through and this feature lands.

I'm happy either way. Your call.

workmanw:
@jamesarosen Honestly, after thinking about it, I do think we should hold. This merge is not prudent and once `isHtmlSafe` lands, will have to commit again. I'm going to close this and I'll take it back up in the future.

Issue #377, opened 2016-06-13
RFC: allow more htmlSafety control


jamesarosen:
Currently (in v4.x), ember-i18n supports the following translation:

``` js
hello: "Hello, <em>{{name}}</em>"
```

If you use this in Handlebars:

``` hbs
{{t "hello" name=user.name}}
```

or in JavaScript:

``` js
hello: Ember.computed('i18n.locale', function() {
  return this.get('i18n').t('hello', { name: this.get('user.name') });
})
```

the result is the same: ember-i18n runs `escapeExpression` on `user.name`, then wraps the whole translation in `htmlSafe`. The first operation prevents XSS attacks. The second prevents overescaped strings -- users seeing _Hello, <em>Vijay</em>_ on the page.

Several people have submitted various issues regarding this escaping-and-safety logic: #280, #374. I believe the best approach to solve this is as follows:

In v4.x, allow calls to `i18n.t` to ask for the `htmlSafe` call to be omitted on a case-by-case basis:

``` js
const i18n = this.get('i18n');
const hello = i18n.t('hello', { name: this.get('user.name'), __htmlSafe: false });
// hello is a plain string, though `user.name` is still escaped
```

Also in v4.x, add a mechanism for users to turn off these `htmlSafe` calls in JavaScript-land globally:

``` js
initialize(application) {
  application.lookup('service:i18n').set('tMarksSafe', false);
}
```

By default, `tMarksSafe` is `undefined`, which is functionally equivalent to `true`, but prints a deprecation warning. Valid overrides are `true` and `false`.

Also in v4.x, add an _escape hatch_ at the helper level in case application developers want plain strings there -- for example, when composing helpers:

``` hbs
{{capitalize (t 'hello' name=user.name __htmlSafe=false)}}
```

In v5.x, remove support for `i18n.Options.htmlSafe` and `tMarksSafe`. All JavaScript-land calls will _not_ mark the string as safe. Helper invocations will wrap in `htmlSafe` by default, but still support `__htmlSafe=false` for composing.

jamesarosen:
@abfx @rainerfrey @workmanw @cfator @gossi I'd love your thoughts on this.

workmanw:
[...]

PR #381, opened 2016-06-28
Reserve locale and htmlSafe attributes

jasonmit:
Based on our discussion over slack /cc @jamesarosen 

This is for handling deprecations ahead of 5.x

Loosely related to #377

### DIFF
--- a/tests/unit/i18n-t-test.js
+++ b/tests/unit/i18n-t-test.js
@@ -1,10 +1,14 @@
 import Ember from 'ember';
 import { moduleFor, test } from 'ember-qunit';
 
-const { run } = Ember;
+const { run, deprecate:originalDeprecate } = Ember;
 
 moduleFor('service:i18n', 'I18nService#t', {
-  integration: true
+  integration: true,
+
+  afterEach() {
+    Ember.deprecate = originalDeprecate;
+  }
 });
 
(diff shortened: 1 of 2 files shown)


tests/unit/i18n-t-test.js
```diff
@@ -36,6 +40,30 @@ test('returns "missing translation" translations', function(assert) {
...
+    }
+  };
+
+  service.t('not.yet.translated', { htmlSafe: true });
+  assert.equal(deprecations, 1);
+
+  service.t('not.yet.translated', { locale: true });
+  assert.equal(deprecations, 2);
+
+  service.t('not.yet.translated');
```
jamesarosen: Maybe `service.t('not.yet.translated', { some: 'other key' })`

jasonmit: Sounds good, I'll add one that includes interpolation

jasonmit: All done, it's on line 63 to catch both w/ and w/o data attribute.

jamesarosen:
+1

jasonmit:
@jamesarosen once merged I'll start on the second part of what we discussed, supporting inlining a locale when looking up a translation.

Issue #383, opened 2016-07-09
pluralized translations fail when count is undefined


jamesarosen:
Given translations

``` js
catCount: {
  one: 'One cat',
  other: '{{count}} cats'
}
```

and this template:

``` hbs
{{t 'catCount', count=numberOfCats}}
```

if `numberOfCats` is 0 or 1 or 1214212342, it works. If it's `undefined`, it shows `missing translation: catCount`. I believe the problem is that the translator only looks for `catCount.one` or `catCount.other` if `count` is _present_, rather than if the _key is passed_.

alexmiddeleer:
Would this cause "missing translation" to appear while a page is loading? Is there an elegant workaround for this that you would recommend? Thanks :)

lcpriest:
I have a default-to helper that I use for this:
```
import Ember from 'ember';
function defined(value) {
  return value !== null && value !== undefined && value !== "";
}

export function defaultTo([value, value1, value2]) {
  if (defined(value)) {
    return value;
  } else if (defined(value1)) {
    return value1;
  } else {
    return value2;
  }
}

export default Ember.Helper.helper(defaultTo);
```

`{{t 'csvs.columns.map' count=(default-to columns.length 0)}}`

jamesarosen:
jamesarosen/ember-i18n has been deprecated in favor of [ember-intl](https://github.com/ember-intl).

Issue #332, opened 2015-11-05
Default to count 1 if not provided


Fed03:
Hi, I'm using this awesome addon and it works great. 
But I think it can be improved in one thing.

Given this

```
// translation.js
"pluralizableWord": {
  "one": "singular",
  "other": "plural"
}
```

I think it would be a better UX if when I'm doing `{{t "pluralizableWord"}}` it would spit out `"singular"`.
In other words, if a localization string is marked for pluralization but count is not provided use the `"one"` version.

let me know what do you think about it. Maybe I could help with a PR if you find it apprpriate

jamesarosen:
The problem with this idea is that this is a totally valid set of translations:
```js
{
  pancakes: 'Some number of pancakes',
  'pancakes.one': 'One pancake',
  'pancakes.other': '{{count}} pancakes'
}

t('pancakes', { count: null }); // 'Some number of pancakes'
```

jamesarosen:
jamesarosen/ember-i18n has been deprecated in favor of [ember-intl](https://github.com/ember-intl).

PR #384, opened 2016-07-14
Non deterministic behavior fixed for plural forms and missing translations

borissuska:
Hi, we found a bug when plural forms are used. When you define plural form like this

``` javascript
  pluralForm: function(count) {
    if (count === 0) { return 'zero'; }
    if (count === 1) { return 'one'; }
    return 'other';
  }
```

and then use translation like this

``` javascript
  'pluralized.short-translation': {
    one: 'One Click',
    other: '{{count}} Clicks'
  },
```

It behaves non-deterministic when you're using count=0.

In case that translation never been used with any count value it returns Missing translation. In case that translation was previously used with count other than 0 it will return for count=0 last used translation. Which is sometimes "One Click", sometimes "0 Clicks".

I found a problem in Locale.findTranslation function and I added defaultPluralForm into config to determine which key should be used in case that pluralForm return value is missing in translation.

I also added test to test this non-deterministic behavior and test to test count=undefined feature (issue #332 and #383). I added sk and cs locale configuration too. I fixed phantomjs ES6 errors too.


jamesarosen:
If the `pluralForm` function for a locale allows for `zero`, then the translations for that locale MUST define a `zero` form. I definitely want to fix the nondeterminism, though. ember-i18n should _always_ report a missing translation for these cases.

Regarding your `defaultPluralForm` solution:

If `n` is a number, it is guaranteed to fall into one of the locale's pluralization rules. If it's _not_ a number, pluralization doesn't really make sense. Your solution would emit something like `"undefined dogs"` when `count` is `undefined`. That seems like a bad experience for end-users, though I could see how it would help programmers find where they're assuming something is a number, but it's not yet loaded.

Thank you for finding this bug and describing it so clearly :)

borissuska:
You're right, I removed `defaultPluralForm` because it disables notification about missing translation. Now I'm not sure what is correct behavior when you got previous pluralForm definition and following translation string:

```javascript
'pluralized.translation': '{{count}} Clicks'
```

Is this an invalid translation definition and for `{{t "pluralized.translation" count=100}}` it should return return "Missing translation" message or is it valid and for any count value it should return "... Clicks" string?

For now it returns "Missing translations: pluralized.translation" because of `other`, `one` and `zero` plural keys are not defined. I'm not sure if it is correct because of backward compatibility. Should I return "... Clicks" string in this case now? If yes, I'll add deprecation message too that developer have to define all locale plural forms for the translation key.

jamesarosen:
```js
'pluralized.translation': '{{count}} Clicks'
```
is an ill-defined translation. It should be at minimum
```js
'pluralized.translation.other': '{{count}} Clicks'
```

borissuska:
As I wrote before this was implemented as you expect.

Now, I improved Missing translation string. If you're looking for a pluralized.translation message with count parameter set, it will return "Missing translation: pluralized.translation.zero", "Missing translation: pluralized.translation.one" or "Missing translation: pluralized.translation.other" depending on pluralForm function for current locale and count value.

addon/services/i18n.js
```diff
@@ -53,7 +53,7 @@ export default Parent.extend(Evented, {
     const count = get(data, 'count');
 
     const translation = locale.findTranslation(makeArray(key), count);
-    return typeOf(translation.result) !== 'undefined';
+    return typeOf(translation.result) !== 'undefined' && !translation.result._isMissing;
```
jamesarosen: Can you rebase on top of master?

jamesarosen:

Issue #256, opened 2015-06-24
Add support for translation fallback


jcbvm:
## Problem

Currently if a user has multiple locale translations defined, it has to define every translation key for each locale, even if some translations may be the same in a locale chain (i.e. en_US -> en).
## Goal

Add the ability to omit translations keys in translations which are or may be defined in other translations files within the same locale chain.
## Proposal

When a translation is being looked up via `t('somekey')` and the result is undefined, we need to track down if the translation is defined in one of the other translations in the current locale chain. This means that we have to lookup the current locale chain and see if there exists a translation in one of the parent locales (i.e. when a translation is not defined in 'en_US', we have to look if the translation exists in 'en', if so we return that translation).

Some key notes:
- Fallback will occur on a string-by-string basis and not if the whole locale is undefined. This way we can omit only certain translation keys.
- When we walk the chain, undefined locales should be skipped.
- When we walked the whole chain and could not find the translation, the current default behavior of an undefined translation should be applied.
## Considerations

It would be possible to let the defaultLocale be the endpoint of the chain, so if the translation cannot be found in the current locale chain, we will finally fallback to the defaultLocale. I'm not sure yet if this  is desirable. 

If there are plans to implement this, the defaultLocale defined in the ENV config should be stored and never be changed when setting the locale property.
## Alternative Approach

An alternative approach would be to merge all translations within one locale chain. This requires to merge translations when initializing the translations and also when new translations are being added via `addTranslations`.

Some key notes:
- The order is important, a more specific locale should overwrite translations of a less specific locale.

Drawbacks:
- It is less efficient, because we have to rebuild the whole translation object when new locales are added, especially when adding a locale in the middle of the current locale chain.

jamesarosen:
Would you mind writing this up more fully in RFC format? Some questions you might want to address:

 * Under what conditions could the user get into a state where the app is asking for translations in a locale that it (the app) doesn't define?
 * Would there be only one fallback or a chain?
 * How would fallbacks (e.g. `nl` -> `en`) work with locale variants (e.g. `en-US` -> `en`)?
 * Would fallback occur on a string-by-string basis or only if the whole locale is undefined? 
 * How would this work with the [multiple packages proposal](https://github.com/jamesarosen/ember-i18n/issues/255)?

jamesarosen:
> Add the ability to omit translations keys in translations which are or may be defined in other translations files within the same locale chain.

[...]

PR #386, opened 2016-07-27
default translations and access to Locale

duanemay:
Added 2 features:
- the ability to have a common default translations file located in app/locales/translations.js 
  I use this for product names, and language names that transcend the locale...
  And therefore I don't want to define it in every base language.
- expose Locale so that I can access translations for a certain language from the code. This is useful in the missing-message.js

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,10 @@
+## 4.2.2
+
+ * set `isLocalizationFramework` for other addons
+ * change `Stream` import to support latest Ember (even though
+   ember-i18n doesn't _rely_ on `Stream` for Ember 1.13+, it needs
+   to _import_ it)
+
 ## 4.2.1
 
  * use `Ember.assign` instead of `Ember.merge` if available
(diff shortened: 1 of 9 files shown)


jamesarosen:
#256 has some good discussion on the fallback/default locale. I think the approach we want to take is to be able to do `i18n.set('defaultLocale', 'es')` and have all locales fall back to `es/translations.js`.

#312 has some good discussion on providing an override locale for one translation. https://github.com/jamesarosen/ember-i18n/pull/381 has a prerequisite for the feature. It involves a reserved key name, which is a breaking change, so this can't go out (by default) until 5.0, though we can add a flag to opt-in early.

toranb:
@duanemay any chance this might still be on your radar (knowing you kicked this off back in the summer of 2016) ?

toranb:
@jamesarosen did this "breaking change" you mention above land in 5X? Does that unlock anything if someone (like me?) decides to take this back on at some point?

duanemay:
@toranb I am no longer in a position where we are using ember, so by all means, you can take it on if you wish.

toranb:
@duanemay sounds like a plan! Care to close this PR? I'll open mine here in a few ;)

toranb:
https://github.com/jamesarosen/ember-i18n/pull/478

^I just threw this PR up to help us land a solution for default locale fallback ;)

PR #390, opened 2016-08-25
exists returns true when checked twice

nerdyworm:
Hey!

We've been hitting this issue which causes an infinite loop when trying to do fallback lookups.  :cry: 

Basically, when you hit a missing translation the system defines a translation with the _isMissing flag. Exists needs to check that flag.

Cheers!


jamesarosen:
Love it! +1

PR #391, opened 2016-08-26
ignore .ember-cli

jamesarosen:
Each developer may want their own custom port and TLS settings. We provide an example file.


jamesarosen:
@rwjblue @stefanpenner what do you think of this

### Use-Case

My Ember app needs to run with TLS locally, so it has `ssl`, `ssl-key`, and `ssl-cert` set in its `.ember-cli` with a self-signed certificate that I've installed in the OS's keychain. That wouldn't have _any_ bearing on this project... _except_ when I try to go to `http://localhost:<any port>`, the OS upgrades me to `https://`. That means _all_ of my ember-cli projects need to use TLS. (This may be related to HSTS, but I'm not certain.)

rwjblue:
Yep, I generally like to exclude `.ember-cli` in my addons and include it in apps (picking a non-standard port for all devs on the app).

PR #392, opened 2016-08-26
Include ember 1.12 support only on 1.12

jamesarosen:
Use ember-cli-version-checker and broccoli-funnel to selectively include
legacy- or non-legacy code in the build depending on the Ember version. For Ember 1.12.x, include `addon/legacy/*` and `app/initializers/ember-i18n-legacy-helper.js`. For Ember 1.13+, include `addon/helper.js` and `app/helpers/t.js`.

 * `addon/helper.js`, the Ember 1.13+ version, can rely on `Ember.Helper` existing
 * `addon/legacy-helper.js` -> `addon/legacy/helper.js`
 * `addon/stream.js` -> `addon/legacy/stream.js`
 * `app/instance-initializers/ember-i18n.js` -> `addon/legacy/initializer.js`, exported to `app/initializers/ember-i18n-legacy-helper.js`
 * remove `app/initializers/ember-i18n.js` as Ember 1.13+ doesn't require any custom initialization

### Benefits

 1. smaller app build sizes since each app needs only the Ember 1.12 or the Ember 1.13+ version
 2. don't have to keep updating the `Stream` import as modern Ember versions change internals (since we only use it on 1.12, which is locked)


jamesarosen:
@rwjblue @blimmer what do you think of this approach?

blimmer:
I'm a big fan of this approach. I actually don't have a 1.12 app anymore, and I think it'd be fine to freeze / not add new features for this old of an app. Most addons are 1.13+ now barring some very old examples.

jamesarosen:
This addon will definitely drop ember 1.12 when it goes 5.0 :)

jamesarosen:
Oh no! `ember beta` doesn't satisfy `ember >= 1.13.0` :(

jamesarosen:
Blocked on rwjblue/ember-cli-version-checker#13

rwjblue:
@jamesarosen - I'm sorry, I'm unsure how this is blocked by that. Can you explain?

index.js
```diff
@@ -1,8 +1,42 @@
...
 
 module.exports = {
   name: 'ember-i18n',
+  isLocalizationFramework: true,
 
-  isLocalizationFramework: true
+  config: function() {
+    var checker = new VersionChecker(this);
+    var dep = checker.for('ember', 'bower');
+    this.hasEmberHelper = dep.satisfies('>= 1.13.0');
```
rwjblue: Prereleases do not satisfy this range:

```
# node
> semver = require('semver')
> semver.satisfies('2.9.0-beta.1', '>= 1.13.0')
false
```

You should use `this.hasEmberHelper = !dep.lt('1.13.0');` for this.

rwjblue:
🎉

PR #395, opened 2016-08-30
Use `Ember.String.htmlSafe` instead of `new Ember.Handlebars.SafeString`.

workmanw:
Using `Ember.Handlebars.SafeString` is deprecated in Ember 2.8. See: http://emberjs.com/deprecations/v2.x/#toc_use-ember-string-htmlsafe-over-ember-handlebars-safestring

A few notes. `Ember.String.htmlSafe` has existed since pre-1.0.0 so there should no concern of backwards compat. Second, the polyfill for `Ember.String.isHtmlSafe` is only used by the tests so it only needs to be a `devDependency`.

### DIFF
--- a/addon/utils/i18n/compile-template.js
+++ b/addon/utils/i18n/compile-template.js
@@ -1,6 +1,6 @@
 import Ember from "ember";
 
-const SafeString = Ember.Handlebars.SafeString;
+const htmlSafe = Ember.String.htmlSafe;
 const get = Ember.get;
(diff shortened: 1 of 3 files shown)


jamesarosen:
I'm glad this deprecation has been sorted out :)

+1

workmanw:
Ditto! There is one more set of deprecation fixes coming from me for Ember 2.8. Perhaps after that you could cut a new release.

PR #396, opened 2016-08-30
Added an `until` property when calling `Ember.deprecate`

workmanw:
This resolves deprecation warnings seen with Ember >= 2.2 and ~~ember-i18n >= 4.2.2~~ ember-i18n > 4.2.2. 

This deprecate was originally added via #381. In that PR it was stated the deprecation was targeted at 5.x. So I made the `until` value reflect that (see diff).


jamesarosen:
+1

workmanw:
Actually just to clarify, this error only happens with master, not 4.2.2. But this should be merged before cutting any new releases and it is backwards compat.

workmanw:
@jamesarosen Thanks for the create turnaround time! 🎉

PR #397, opened 2016-08-30
Fixing new issue with Ember 2.9

workmanw:
With Glimmer2, Ember.Helpers will receive their `namedArgs` as an EmptyObject. This created a problem using `hasOwnProperty` in the service. This is a simple work around.

See: https://github.com/emberjs/ember.js/issues/14161


jamesarosen:
+1

PR #399, opened 2016-09-09
Use `semver.lt` instead of `semver.satisfies`.

rwjblue:
`semver.satisfies` will never match prerelease versions for any version other than the one being tested.

We have to use `gt` and `lt` for that (but ember-cli-version-checker only exposes `lt` IIRC).

**NOTE:** This is a rebase of https://github.com/jamesarosen/ember-i18n/pull/392 with a one line fix.


rwjblue:
OK, this failed CI due to the base branch not including the changes from https://github.com/jamesarosen/ember-i18n/pull/397 yet, but it does *not* have the error about `ember-htmlbars/streams/stream` (and only has 8 failures instead of all the tests failing).

rwjblue:
* Updated target branch of this PR to target master
* Rebased #392's changes along with this change
* Triggered CI run
* All CI builds are green.

jamesarosen:
Merged these changes into #392. Thanks :)

rwjblue:
Awesome, thank you!

PR #406, opened 2016-09-14
Check ember version during init callback

### LINKED ISSUE #401: Stream not available in 2.7.0  (v4.3.0)
I was just upgrading a private add-on to 2.7.0 and found this broke it due to legacy files coming in (stream).

`https://github.com/jamesarosen/ember-i18n/blob/master/index.js#L10`

`config(){...}` appears to be called multiple times when I run `ember s`, sometimes 
`this.hasEmberHelper` is `undefined` other times its `true`.

moving it to `init(){...}` fixes it.

I'm not too familiar with add-on callbacks yet, any ideas why `init` works but not `config`?

luqman:
Fixes `Could not find module ember-htmlbars/streams/stream` error. Config callback caused `hasEmberHelper` to be set to `undefined`.

Fixes #401

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,11 +7,12 @@ module.exports = {
   name: 'ember-i18n',
   isLocalizationFramework: true,
 
-  config: function() {
+  init: function() {
+    this._super.init && this._super.init.apply(this, arguments);
+
     var checker = new VersionChecker(this);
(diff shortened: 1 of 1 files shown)


jamesarosen:
SGTM

PR #410, opened 2016-10-14
Do not observe computed properties that are not consumed

npars:
- Ember docs state that computed properties that are not consumed will not trigger observers, unfortunately this is not true for readOnly properties due to a current bug in ember emberjs/ember.js#9265
- Update the i18n helper to not observe a readOnly computed property for when this issue is eventually fixed


jamesarosen:
+1. Thanks!

PR #418, opened 2016-11-08
Extract getPluralKey. Allow redefine Locale props on instantiate from service

vvscode:


addon/services/i18n.js
```diff
@@ -94,7 +100,8 @@ export default Parent.extend(Evented, {
 
   _locale: computed('locale', function() {
     const locale = this.get('locale');
+    const localeProps = this.get('localeProps');
 
-    return locale ? new Locale(this.get('locale'), getOwner(this)) : null;
+    return locale ? Ember.merge(new Locale(this.get('locale'), getOwner(this)), localeProps) : null;
```
jamesarosen: This interface isn't very obvious to me. What problem are you trying to solve? If you want to customize how a `Locale` is built, maybe we should have

```js
// @public
buildLocale(id) {
  return new Locale(id, getOwner(this));
},

_locale: computed('locale', function() {
  const locale = this.get('locale');
  return locale ? this.buildLocale(locale) : null;
}
```

vvscode:
@jamesarosen is it ok?

jamesarosen:
You have some typos ("tranclation"). Can you also write a little about the use-case for changing the way plural keys are constructed? I ask because in v5.0, I'm planning on moving all the compilation to the build process and many of these APIs will cease to exist.

PR #422, opened 2016-12-07
Updated the usage of `ember-getowner-polyfill` to fix deprecation

workmanw:
An update to the `ember-getowner-polyfill` caused deprecations to be spewed in my builds. Fixing that. See: https://github.com/rwjblue/ember-getowner-polyfill/issues/5 .

### DIFF
--- a/addon/services/i18n.js
+++ b/addon/services/i18n.js
@@ -1,10 +1,9 @@
 import Ember from "ember";
-import getOwner from 'ember-getowner-polyfill';
 import Locale from "../utils/locale";
 import addTranslations from "../utils/add-translations";
 import getLocales from "../utils/get-locales";
 
-const { assert, computed, get, Evented, makeArray, on, typeOf, warn } = Ember;
+const { assert, computed, get, Evented, makeArray, on, typeOf, warn, getOwner } = Ember;
 const Parent = Ember.Service || Ember.Object;
 
 // @public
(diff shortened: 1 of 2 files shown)


jamesarosen:
We should change the dependency in `package.json` to `>=1.1.0` or `^1.1.0` to make sure we get the new behavior?

rwjblue:
@workmanw:

* There should only be one deprecation from all usages of ember-getowner-polyfill (since the deprecation is done when the module is evaluated), can you confirm?
* Can you update the the `package.json` to require at least the latest version (so that we can reliably rely on `Ember.getOwner` having been populated)?

workmanw:
@jamesarosen Yeap! Good catch sir.

---------

@rwjblue 

> There should only be one deprecation from all usages of ember-getowner-polyfill (since the deprecation is done when the module is evaluated), can you confirm?

Yeap, can confirm. The building of 3 ember apps, plus the extra deprecation for _"you must provide `until` in options"_ made it feel like more than it was.

> Can you update the the package.json to require at least the latest version (so that we can reliably rely on Ember.getOwner having been populated)?

Yeap. Done.

jamesarosen:
LGTM

locks:
:shipit: 😀

lolmaus:
Can you please release this? :bow:

jamesarosen:
Yep. As soon as we figure out #423.

jamesarosen:
Released in v4.4.0

lolmaus:
:bow:

MoxyFruzy:
Can 4.4.0 be released to nom please. The latest available package is 4.3.2.

PR #431, opened 2017-01-24
Use owner.factoryFor is present

cibernox:
This removes deprecations and should have a very positive perf outcome in Ember 2.12

### DIFF
--- a/addon/services/i18n.js
+++ b/addon/services/i18n.js
@@ -69,7 +69,8 @@ export default Parent.extend(Evented, {
 
   // @private
   _initDefaults: on('init', function() {
-    const ENV = getOwner(this)._lookupFactory('config:environment');
+    let owner = getOwner(this);
+    let ENV = owner.factoryFor('config:environment').class;
 
     if (this.get('locale') == null) {
       var defaultLocale = (ENV.i18n || {}).defaultLocale;
--- a/addon/utils/add-translations.js
+++ b/addon/utils/add-translations.js
@@ -4,7 +4,8 @@ const assign = Ember.assign || Ember.merge;
 
 export default function addTranslations(locale, newTranslations, owner) {
   const key = `locale:${locale}/translations`;
-  var existingTranslations = owner._lookupFactory(key);
+  let factory = owner.factoryFor(key);
+  let existingTranslations = factory && factory.class;
 
   if (existingTranslations == null) {
     existingTranslations = {};
--- a/addon/utils/locale.js
+++ b/addon/utils/locale.js
@@ -29,8 +29,7 @@ Locale.prototype = {
       if (this.rtl === undefined) { this.rtl = config.rtl; }
       if (this.pluralForm === undefined) { this.pluralForm = config.pluralForm; }
     });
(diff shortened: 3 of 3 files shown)


workmanw:
Just looking at these changes, it's not obvious to me where the performance improvement comes from. I was wondering if you could elaborate on that or maybe point me toward something I could read.

cibernox:
The factoryFor method seems to be way faster, according with the release notes of 2.12.
And even if it wasn't, the other method is deprecated.

rwjblue:
The factory for RFC outlines the perf benefits. The short version is that Ember no longer has to `.extend` every factory a second time (which we have done for a long time to ensure it gets the proper injections).

workmanw:
Oh awesome! Thank you @cibernox @rwjblue. I read the PR, but it wasn't clear to me there. I will definitely go reread the RFC. :+1:

cibernox:
@bwbuchanan Added

addon/utils/locale.js
```diff
@@ -115,19 +133,36 @@ function getFlattenedTranslations(id, owner) {
     assign(result, getFlattenedTranslations(parentId, owner));
   }
 
-  const translations = owner._lookupFactory(`locale:${id}/translations`) || {};
-  assign(result, withFlattenedKeys(translations));
+  let translations;
+  if (owner.factoryFor) {
+    let maybeTranslations = owner.factoryFor(`locale:${id}/translations`);
+    translations = translations && translations.class;
```
bwbuchanan: Should be `translations = maybeTranslations && maybeTranslations.class;`

bwbuchanan:
Looks good. My app's tests are green.

cibernox:
Actually mine had so many deprecations that tests broke because travis cannot handle a log that big.

addon/services/i18n.js
```diff
@@ -69,7 +69,13 @@ export default Parent.extend(Evented, {
 
   // @private
   _initDefaults: on('init', function() {
-    const ENV = getOwner(this)._lookupFactory('config:environment');
+    let owner = getOwner(this);
+    let ENV;
+    if (owner.factoryFor) {
+      ENV = owner.factoryFor('config:environment').class;
```
rwjblue: In general, we should return the object returned by factoryFor (not the .class property)

addon/utils/add-translations.js
```diff
@@ -4,7 +4,13 @@ const assign = Ember.assign || Ember.merge;
 
 export default function addTranslations(locale, newTranslations, owner) {
   const key = `locale:${locale}/translations`;
-  var existingTranslations = owner._lookupFactory(key);
+  let existingTranslations;
+  if (owner.factoryFor) {
+    let maybeTranslations = owner.factoryFor(key);
+    existingTranslations = maybeTranslations && maybeTranslations.class;
```
rwjblue: Remove .class

cibernox:
**Don't merge** this yet. I've realized that I misunderstood how `factoryFor` is supposed to be used. I'll update this PR.

PR #433, opened 2017-01-28
Use VersionChecker's forEmber to support ember-source

cesarizu:
This PR updates ember-cli-version-checker to support newer ember versions that use ember-source.


locks:
Thank you!

mriska:
Thanks for fixing. Any ideas when a new version will be released? I currently have one project that builds locally but fails in Travis without this fix.

PR #435, opened 2017-02-13
Use more ES6 syntax (thx ember-watson)

cibernox:


blueprints/locale/index.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   description: 'Generates new i18n locale directory with config and translations files.',
 
-  beforeInstall: function(options) {
```
jamesarosen: This requires Node 6, doesn't it?

index.js
```diff
@@ -7,15 +7,15 @@ module.exports = {
   name: 'ember-i18n',
   isLocalizationFramework: true,
 
-  init: function() {
+  init() {
```
jamesarosen: This would also require a fairly recent version of node

rwjblue: It is Node 4+ (and Node 0.12 is EOL)...

blueprints/locale/index.js
rwjblue: Node 4 or higher

jamesarosen:
+1

PR #437, opened 2017-02-15
Make pluralForm() for Turkish always return `ONE`

### LINKED ISSUE #436: Turkish pluralForm should always return `ONE`
According to [this resource](http://learnturkish.pgeorgalas.gr/SingularVsPluralSetEn.asp):

> When you use a number or any word that indicates quantity before a noun, use the singular form of the noun. 

At present, the `pluralForm` in [`tr.js`](https://github.com/jamesarosen/ember-i18n/blob/a50133e9d73fbe010f74aca3936d834c482fedc7/addon/config/tr.js) does the same thing as [`zh.js`](https://github.com/jamesarosen/ember-i18n/blob/a50133e9d73fbe010f74aca3936d834c482fedc7/addon/config/zh.js), which is to always return `OTHER`. Instead, it should always return `ONE`.

nelstrom:
Fixes #436

### DIFF
--- a/addon/config/tr.js
+++ b/addon/config/tr.js
@@ -1,2 +1,9 @@
-import zh from './zh';
-export default zh;
+import { ONE } from './constants';
+
+export default {
+  rtl: false,
+
+  pluralForm() /* n */{
+    return ONE;
+  }
+};


cibernox:
Twitter confirmation https://twitter.com/ahakanergun/status/831930758508007429

Approved

jamesarosen:
The [CLDR for tr](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#tr) says the rule should actually be the same as English:
```js
if (n == 1) return ONE;
return OTHER;
```

PR #445, opened 2017-03-31
fixes getLocales returning duplicates

huafu:
`getLocales()` is returning duplicates in case you have addons with translations in your app. Then you end-up having duplicates in the `i18n.locales` service's property, and since it's the property used in the example to create the language switch, you endup with many times the same language in that switch.


jamesarosen:

PR #447, opened 2017-04-10
Add support for automatically choosing different i18n keys for different countries

inez:
Example: if countryCode is set to RU then lookup keys in order: <key>.RU, <key>.default, <key>

### DIFF
--- a/addon/services/i18n.js
+++ b/addon/services/i18n.js
@@ -13,6 +13,17 @@ export default Parent.extend(Evented, {
   locale: null,
 
   // @public
+  countryCode: Ember.computed({
+    set(key, value) {
+      if(Ember.isBlank(value) || value.length !== 2 || value !== value.toUpperCase()) {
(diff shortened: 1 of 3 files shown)


toranb:
Whoa ... so this PR has just hung (green) since Apr 10th 2017 without any activity ?!??

@inez can you confirm that what I've done w/ missing message accomplishes the same goal you are trying to add here?

note: obviously mine is way more of a hack atm ;) and doesn't support a full template compile

^if so ... what do we need to land this? I've been asking around and it seems this would be a no-brainier to add #am_i_rite

toranb:
Ah, disregard that comment seems this is for defaults within a single locale ...

@snewcomer @jamesarosen you guys know of anyone who has tried to add support for a "default" fallback like I did in that gist above?!??

toranb:
ah looks like #386 is what I was after! I'll skip out of this PR and read up on that one

jamesarosen:

PR #451, opened 2017-07-06
Add danish language

cibernox:


cibernox:
If test pass I'll merge this. The plurals of my app are broken in danish.

PR #454, opened 2017-07-31
Update ember-getowner-polyfill

jrjohnson:
Major version bump removes the deprecated code from 1.x. This addon
doesn't rely on the removed import syntax, it pulls getOwner off of the
Ember object (#422) so this is an easy change.


locks:
Thank you!

PR #457, opened 2017-08-11
Replace jshint with eslint

aureliosaraiva:


jamesarosen:

PR #461, opened 2017-10-25
ember.2.16

majnikov:
Updating for ember v2.16
[according to this](https://www.emberjs.com/blog/2017/10/11/ember-2-16-released.html)


majnikov:
sorry, mistaken PR

PR #462, opened 2017-10-27
Delete LICENSE.md

### LINKED ISSUE #452: Which license file is valid?
I see there are 2 licenses in the root of this project. The MIT license appears to be newer but there's an Apache-2.0 license in there, which is also referenced from `package.json`.

Which license is correct, please? If it's just one of them would it be possible to remove the other and/or update `package.json`? Our org doesn't seem to like license ambiguity :)

sandstrom:
As requested in https://github.com/jamesarosen/ember-i18n/issues/452#issuecomment-314517389

closes #452

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,9 +1,0 @@
-The MIT License (MIT)
-
-Copyright (c) 2017
-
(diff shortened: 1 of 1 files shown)


sandstrom:
Would you like to merge it? 😄

PR #467, opened 2017-12-05
Use Ember.Object.create instead of .extend().create()

### LINKED ISSUE #465: helper creates Ember.Object.extend().create()
https://github.com/jamesarosen/ember-i18n/blob/df616c0b1eb4f25b27d78cbc220e9a6a0b863d4d/addon/helper.js#L6-L11

The `t` helper is extending `Ember.Object` and then creating an object on every usage of the helper. I'm not *really* sure what is going on, but I found a strange bug where `Object.assign()` does't work as expected on `Ember.Object.extend().create()`.:

```js
Object.assign({}, Ember.Object.extend({foo: 'bar'}).create())
#=> {}

Object.assign({}, Ember.Object.create({foo: 'bar'}))
#=> {foo: 'bar'}
```

In my code, I'm using this Object.assign() to clone the data that comes into this method and it's causing strange problems.

I *think* we could use `Ember.Object.create()` directly in this method and then go chase down the underlying problem in Ember separately?

mehulkar:
Fixes #465

### DIFF
--- a/addon/helper.js
+++ b/addon/helper.js
@@ -3,12 +3,12 @@ import Ember from "ember";
 const { get, inject, Helper, Object: EmberObject, observer } = Ember;
 
 function mergedContext(objectContext, hashContext) {
-  return EmberObject.extend({
+  return EmberObject.create({
     unknownProperty(key) {
       const fromHash = get(hashContext, key);
       return fromHash === undefined ? get(objectContext, key) : fromHash;
     }
-  }).create();
+  });
 }
 
 export default Helper.extend({


locks:
CI is complaining about JSHint 🤔

mehulkar:
hmm, yeah i saw that locally too but assumed that the problem already existed or was local to my machine. I'll look into it again, brb

mehulkar:
yeah, looks like it fails on master also, and the error is for checking if a global is available. Not sure if there's a good way to fix this without ignoring the usage, since even in Ember, `find` (and maybe even `QUnit`?) is a global right now.

nathanhammond:
Worth noting, this adds the `unknownProperty` as an owned, enumerable property to the object being consumed (which is why it is documented as `extend().create()` upstream. (You would typically not want to expose this as such.)

However, doing as we've done creates a new shape on every usage. The `i18n.t()` function will go megamorphic immediately and stick at baseline, which is super not ideal for something as hot as the `t` helper.

I'd actually recommend considering a modification to https://github.com/jamesarosen/ember-i18n/blob/df616c0b1eb4f25b27d78cbc220e9a6a0b863d4d/addon/services/i18n.js#L23 to make it possible to pass both objects forward.

```javascript
t(key, ...contextObject) {
  // for each contextObject
    // call get, but hoist the check for `unknownProperty`. https://github.com/emberjs/ember.js/blob/d5f57aad5f5a9b8d62322108ac4571d70ca3a00c/packages/ember-metal/lib/property_get.js#L65
    // use the first resolved value.
  // Do the rest of the current function.
}
```

No additional objects would be created, nothing needs to be modified, and the API shape changes in  a backwards-compatible manner.

nathanhammond:
I misunderstood how far this would have to be passed forward:
https://github.com/jamesarosen/ember-i18n/blob/028c41bc89145a959a99caf8334712e26d6ac348/addon/utils/i18n/compile-template.js#L18

I still believe that it makes sense to move it.

jamesarosen:
@nathanhammond you're welcome to break public API in a major version, which we can do any time. And it's much better to break the public API for customizing template compilation (which few users of the library do) than template _invocation_ (which all users do). I still don't understand your argument, but if you can provide the PR, I'm game.

mehulkar:
@jamesarosen the breaking tests seem to have fixed themselves, btw! (i think i just rebased and it worked).

jamesarosen:
@nathanhammond why `...data`? What's the benefit of multiple context objects?

mehulkar:
@jamesarosen any interest in keeping this open for longer? The change seems to be a good one to me, but I'll close this if I don't hear back from you in the next few days.

jamesarosen:
@mehulkar I'm fine with merging it, but I'm about to formally deprecate this library. I'll make you an owner as well.

PR #469, opened 2017-12-19
Update all the things

cibernox:
Update everything to 2.17

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,10 @@
 /coverage/*
 /libpeerconnection.log
 npm-debug.log*
+yarn-error.log
 testem.log
+
+# ember-try
+.node_modules.ember-try/
+bower.json.ember-try
+package.json.ember-try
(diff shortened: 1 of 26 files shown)


.gitignore
```diff
@@ -14,4 +14,10 @@
 /coverage/*
 /libpeerconnection.log
 npm-debug.log*
+yarn-error.log
```
jamesarosen: Is this just from ember-cli's blueprint? Because we're not using Yarn here

cibernox: Yes, it's part of the default blueprint

tests/.eslintrc.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  env: {
+    embertest: true
+  },
+  globals: {
+    t: true,
+    expectTranslation: true
```
jamesarosen: `true` means that test files can override the value, which seems odd. I would expect `false` for both of these.

PR #470, opened 2017-12-20
Add an `/addon-test-support` folder with a new `{{t}}` helper to be u…

### LINKED ISSUE #468: Refactor test helpers
There is several ways in which the test helpers should be improved:

1. The idiomatic place to put test helpers is inside `/addon-test-support`. We should move them. To ease the transition, we should import and reexport with a deprecation from `/test-support` for a while.
2. Afaik `Ember.Test.registerHelper` is not required (I think it doesn't even work!) with the new test syntax proposed in https://github.com/emberjs/rfcs/pull/268 and shipped already in `ember-qunit` 4.2.0. Helpers. Those helpers should be explicitly importable with `import { expectTranslation } from 'ember-i18n/test-support/helpers';`. That means they would also work in integration tests too.
3. Make the helpers not rely on jQuery, as some apps don't use it anymore, and the goal is to not include it by default in future versions of Ember.
4. IMHO, I believe a custom assertion like `assert.hasTranslation('#selector', 'login.button.submit', { count: 4 }, 'Optional assertion description')` would be more idiomatic. Although personally I would use a mix of qunit-dom and the `t` test helper: `assert.dom('#selector').hasText(t('login.button.submit', { count: 4 }), 'Optional assertion description')`

This radar is so I don't forger, but I'll try to work on those points myself over Xmas. But if someone wants to beat me to it, just leave a comment here and do it.

cibernox:
…sed with the new testing syntax

The idea is that everybody using the old approach will be able to continue to do so, 
but those using the new `setupRenderingTest/setupApplicationTest` will be able to just

```js
import { t } from 'ember-i18n/test-support';

//...

  test('Some test', async function(assert) {
    await render(hbs`.....`);
    assert.dom('#header').hasText(t('my.loc.key', { count: 3 }));
  });
```

If this is merged we need to update the wiki page about testing, clarifying that the helper(s) from this new module must be used on the new tests.
I decided not to port `expectTranslation` because I believe it is a no man's land. It's not idiomatic for QUnit users and its not idiomatic for mocha users either.
If approved I think this closes #468


cibernox:
@rwjblue I followed your advice here

test-support/helpers/ember-i18n/test-helpers.js
```diff
@@ -1,15 +1,15 @@
 /* globals QUnit, expect */
 
-import Ember from 'ember';
+import { registerHelper } from '@ember/test';
+import { _t } from 'ember-i18n/test-support/helpers';
```
Turbo87: I think this will be an issue for people that don't have `@ember/test-helpers` yet, as the import above will fail. Either you move `_t()` to a module that has no dependency on `@ember/test-helpers` or you import it using intimate API (`require()`) and wrap that in a try-catch

cibernox: You're right, that will probably fail. I'll move it.

rwjblue:
> import { t } from 'ember-i18n/test-support/helpers';

Thoughts on reexporting in `addon-test-support/index.js` to make the import a bit shorter (and subjectively nicer)?

addon-test-support/helpers.js
```diff
@@ -0,0 +1,7 @@
+import { getContext } from '@ember/test-helpers';
```
rwjblue: FWIW, you can change the import path to `import { getContext } from `ember-test-helpers’` and have support going waaaaaay back. That module is shipped still, and `getContext` has existed for quite some time. The only caveat is that `.owner` won’t exist in older style tests (there you would need `.container` probably?).

I don’t feel strongly at all here, just pointing out some options...

cibernox:
I don't mind. Is there a convention?

rwjblue:
I think (in my head at least) reexporting from index.js of a given tree indicates “public api” and deeply importing from a sub path may or may not be public api. Again, really doesn’t matter much either way.

addon-test-support/helpers.js
cibernox: I think we can leave it this way and tell users this only works with the new tests, and perhaps use it as a carrot for people to update.

cibernox:
I'll export stuff from index then.

addon-test-support/helpers.js
rwjblue: 👍

cibernox:
@jamesarosen Can you make a new release? This does not have any breaking change, so it can be a minor version bump.

jamesarosen:
> Can you make a new release?

I sure can. Let's also add you to the npm owners list!

jamesarosen:
We should also update the testing section of the wiki.

cibernox:
@jamesarosen I already did.

PR #473, opened 2018-01-21
LowerCase Everything

Alonski:
In our app we somehow had translations that were Uppercase or Lowercase.
They should both translate to the same translations but currently this addon doesn't work like that.

Does it make sense to enforce this or to add an option to declare something like this?


jamesarosen:
This is a breaking change and would have to go through the RFC process. What other i18n system does this?

Alonski:
Hey @jamesarosen, 
I understand that this is a breaking change and I don't intend for this to be approved right away without discussion. I tried to override the translation functions but couldn't find a way to do this. Maybe it is possible and you can point it out for me? 

Also this could be a environment flag as well possibly?

jamesarosen:
> Also this could be a environment flag as well possibly?

I suppose we could, but I really don't think this is a concern for the library. If there were two major competing standards about how to write translation keys -- one with mixed-case and the other all lower-case -- then I'd be for a flag. But there aren't. It's just up to the preference of the application developer.

> I tried to override the translation functions but couldn't find a way to do this. Maybe it is possible and you can point it out for me?

This is probably the easiest way to do it:

```js
// app/services/i18n

import I18nService from 'ember-i18n/services/i18n'

export default I18nService.extend({
  t(key, data = {}) {
    return this._super(key.toLowerCase(), data)
  },

  exists(key, data ={}) {
    return this._super(key.toLowerCase(), data)
  }
})
```

Alonski:
That works much better.
I tried to override utils but didn't think to override the service itself.
Thank you very much!
Would you like for me to add a page to the Wiki about this?

jamesarosen:
> Would you like for me to add a page to the Wiki about this?

Not a bad idea!

PR #474, opened 2018-01-24
Update ur.js

jdurand:
Urdu is RTL

### DIFF
--- a/addon/config/ur.js
+++ b/addon/config/ur.js
@@ -1,2 +1,5 @@
 import en from './en';
-export default en;
+
+export default {
+  rtl: true
+};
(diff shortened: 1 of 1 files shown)


addon/config/ur.js
```diff
@@ -1,2 +1,5 @@
 import en from './en';
-export default en;
+
+export default {
+  rtl: true
+};
```
jamesarosen: You'll need some pluralization rules. If you want the ones from English (1 = "one", everything else is "other"), you could do
```js
import en from './en'
export default {
  pluralForm: en.pluralForm,
  rtl: true
}
```

jamesarosen:
Change looks good. Something is wrong with the build.

jdurand:
I think Travis just has a little cold. He'll get there.

jdurand:
Looks like it's a proper flu after all...

jamesarosen:

PR #476, opened 2018-02-08
use textContent instead of innerHTML

snewcomer:
@jamesarosen I think I made a mistake in the original PR.  Maybe `textContent` would be better.  Plus stuff like `&` and such would be parsed differently.

### DIFF
--- a/addon-test-support/-private/assert-translation.js
+++ b/addon-test-support/-private/assert-translation.js
@@ -3,7 +3,7 @@
 export default (function() {
   if (typeof QUnit !== 'undefined' && typeof QUnit.assert.ok === 'function') {
     return function(element, key, text) {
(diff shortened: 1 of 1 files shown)


jamesarosen:
How about `innerText`?

snewcomer:
@jamesarosen Hm.  That is a good question.  What we really want is to emulate `:contains` from the previous version of the helper so users aren't tripped up.  From that article you posted, the only difference I can find is wrt `hidden` elements (`<span>Hello <span style="display: none;">World</span></span>`).  And `innerText` does not extract the text, `jquery` does, and `textContent` does.  So I'm thinking `textContent` may be what we want.  Lmk if you see something else.

Firefox doesn't support `innerText`.  Also saw this wrt innerText:
```
This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.
```

jamesarosen:
`textContent` sounds good to me then. +1

PR #475, opened 2018-02-05
export "expectTranslation" for new acceptance testing system

snewcomer:
https://github.com/emberjs/rfcs/blob/master/text/0268-acceptance-testing-refactor.md

@jamesarosen

PR #477, opened 2018-02-16
replace innerHtml with textContent

snewcomer:
Forgot one.  Possible to get a patch upgrade after this PR! 

Note: used helper on a lg test suite and everything worked great!

### DIFF
--- a/addon-test-support/-private/assert-translation.js
+++ b/addon-test-support/-private/assert-translation.js
@@ -7,7 +7,7 @@ export default (function() {
     };
   } else if (typeof expect === 'function') {
     return function(element, key, text) {
-      const found = !!(document.querySelector(element).innerHTML.indexOf(text) !== -1);
+      const found = !!(document.querySelector(element).textContent.indexOf(text) !== -1);
       expect(found).to.equal(true);
     };
   } else {


jamesarosen:
> Possible to get a patch upgrade after this PR!

I'm wondering whether the change from `:contains` to `textContent` means it has to be a _major_ version bump.

snewcomer:
Hm.  Yeah you may be right.  A user could explicitly import this helper in the old testing API's and it would fail if they haven't migrated to the new testing API's (`@ember/test-helpers` specifically).  However, if users have migrated to the new testing API's, it is safe to assume they have gotten rid of `:contains` selectors...

jamesarosen:
Specifically, I'm thinking of the difference between [before](https://github.com/jamesarosen/ember-i18n/pull/475/files#diff-b37a555760fbeb3f7e4ce23503aad541L21) and [after](https://github.com/jamesarosen/ember-i18n/pull/475/files#diff-8251a387fb0a902445da9f3e5991e6afR6).

snewcomer:
Ah yes.  Anybody still on the old testing API's will see a failure.  I guess that would constitute a major bump!

PR #487, opened 2018-04-30
Stop setting Chinese as the default.

nathanhammond:
We should be using a private use language subtag.

### DIFF
--- a/addon/config/x-i18n.js
+++ b/addon/config/x-i18n.js
@@ -1,0 +1,9 @@
+import { OTHER } from './constants';
+
+export default {
+  rtl: false,
+
+  pluralForm() /* n */{
+    return OTHER;
+  }
+};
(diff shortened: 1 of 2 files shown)


jamesarosen:
I love it

nathanhammond:
Didn't forget, just discovered that doing it in the GitHub UI required me to do it in two commits. 😜

jamesarosen:
> is this going to be a major version update though?

It doesn't feel like a breaking API change to me. I'm open to arguments either way.

mehulkar:
I guess since the implementation is the same as the implementation for `zh`, it's not a breaking change! 

The only case where this might break people is if they've overriden `zh` config specifically, and then encounter a locale that isn't included in this package. (For me, for example, I defined a config for `el-gr`, which is what triggerred @nathanhammond's PR)

PR #490, opened 2018-06-22
remove ancient ember version support

izelnakri:
I think in future releases it is better if we dont support any ember version below ember lts 2.12. This also results in slightly faster builds and less packages to download for current users.


jamesarosen:
This is great, but it technically needs to be a major version bump.

izelnakri:
👍 sounds good

jamesarosen:

PR #491, opened 2018-06-24
Module unification support

izelnakri:
***WARNING:*** this only works for module unification based resolvers, however support for legacy directory(`app/locales`) is trivial.

This includes changes from https://github.com/jamesarosen/ember-i18n/pull/490 so merge that first.

This pull request loads localization files/caches directly from requirejs modules insted of ember containers because container registration logic has changed in the new Module unification resolver. In addition, new ember resolver restrict lookups to only ones defined in the resolver, thus `locale:en/translations` throws unrecognized type `locale`.

I've discovered these issues and made the fix during my migration of a fairly complex ember.js application from `ember-cli` to `mber`.

Furthermore, I highly suggest a rewrite of `addon/utils/locale.js`. Its very hard to read and low quality due to prototypical implementation and its history. Locale unit tests are very simple as well.


jamesarosen:

PR #495, opened 2018-10-19
🔥 deprecate ember-i18n in favor of ember-intl

jamesarosen:


cibernox:
Maybe we should release a new version that prints a deprecation in the console?

lifeart:
nooooo.... :( sad news

cibernox:
@lifeart I know it may seem bad news but it's actually the opposite. Ember will now have a standard package for internationalization. Ember-intl uses the standard Intl API from the browser and it's more powerful.

jamesarosen:
> Maybe we should release a new version that prints a deprecation in the console?

Great idea! I'll open another PR against the 5.x branch.

> sad news

I agree that it's sad, but also with @cibernox that it's for the best. Maybe melancholy?

jamesarosen:
See also https://github.com/jamesarosen/ember-i18n/releases/tag/v5.3.0

PR #498, opened 2019-01-28
Use console-ui for deprecation message

simonihmig:
This gives some better formatting (yellow chalk), but also acknowledges the `--silent` flag of Ember CLI.

The latter does give us some serious problems here, as we parse the test output (in xunit format) in CI, and because of the (non-silent) console output, the output of `ember t` is obviously no valid XML anymore, failing our builds.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,7 +13,7 @@ module.exports = {
     var checker = new VersionChecker(this);
     var dep = checker.forEmber();
     this.hasEmberHelper = !dep.lt('1.13.0');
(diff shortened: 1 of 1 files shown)


mehulkar:
oh wow, I guess @jamesarosen granted me merge access to this repo at one point? I can hit the button, but don't think I can release a new version, so I'll just wait for @jamesarosen.

jamesarosen (approved):
+1

jamesarosen:
Published as 5.3.1

